PR #3, opened 2018-11-13
NATIVE-25: Improve ImageView Modal

NKlug:


src/modules/layout/components/ModalHeader.js
```diff
@@ -56,19 +41,16 @@ class Header extends React.PureComponent<PropsType, StateType> {
   }
 
   render () {
-    const headerTitle = this.getDescriptor().headerTitle || ''
-
     return (
       <BoxShadow theme={this.props.theme}>
         <Horizontal>
           <HorizontalLeft>
             <HeaderBackButton onPress={this.goBack} />
```
maxammann: the null in goBack is acually required by design: https://reactnavigation.org/docs/en/navigation-prop.html#goback-close-the-active-screen-and-move-back

src/routes/image/components/ImageViewModal.js
```diff
@@ -1,13 +1,21 @@
 // @flow
 
 import * as React from 'react'
-import { View } from 'react-native'
+import styled from 'styled-components'
 import { ImageViewer } from 'react-native-image-zoom-viewer'
 import type { NavigationScreenProp } from 'react-navigation'
+import type { ThemeType } from '../../../modules/theme/constants/theme'
+
+// adding a negative margin to also show the image beneath the header
```
maxammann: This is very weird. This should not be necessary.

src/modules/layout/components/ModalHeader.js
```diff
@@ -19,12 +19,6 @@ const HorizontalLeft = styled.View`
...
 `
 
-const Title = styled.Text`
- font-size: 30px;
- color: black;
- margin-left: 10px;
-`
-
 const BoxShadow = styled.View`
   background-color: transparent;
```
maxammann: You can use:

```
const BoxShadow = styled.View`
  background-color: transparent;
  position: absolute;
  z-index: 100;
  top: 0;
  left: 0;
  right: 0;
  height: ${props => props.theme.dimensions.modalHeaderHeight};
`
```

That way the ModalHeader is always transparent and does not take space. You can also name this then TransparentHeader maybe. So it is more reusable :)

src/routes/image/components/ImageViewModal.js
maxammann: See the comment below. It actually is necessary and works as expected :)

src/modules/layout/components/ModalHeader.js
NKlug: Done. This somehow fixed the bug that the image is not displayed beneath the header. :+1:

src/modules/app/components/Navigator.js
```diff
@@ -40,28 +40,42 @@ export const LandingStack = createSwitchNavigator(
...
 export default createStackNavigator(
   {
-    'LandingStack': LandingStack,
-    'ChangeLanguageModal': ChangeLanguageModalContainer,
-    'ModalStack': ModalStack,
-    'MapViewModal': MapViewModal,
-    'ImageViewModal': ImageViewModal
+    'LandingStack': {
+      screen: LandingStack,
+      navigationOptions: {
```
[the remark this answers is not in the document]
NKlug: Done.

PR #4, opened 2018-11-27
NATIVE-17: Display errors

NKlug:


src/modules/common/hocs/withError.js
```diff
@@ -0,0 +1,28 @@
+// @flow
```
maxammann: I think we can move the error handling code into an `error/` module if it has no dependencies to the app module.

src/modules/common/hocs/withError.js
```diff
@@ -0,0 +1,28 @@
...
+
+      if (error) {
+        return <Failure error={new Error(error)} />
+      }
+
+      return <Component {...props} />
+    }
+  }
+
+  // $FlowFixMe Please have a look at this flow warning
```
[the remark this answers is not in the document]
maxammann: I'll look into it.

src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -17,15 +17,15 @@ const mapStateToProps = (state: StateType, ownProps) => {
   const targetCity: string = ownProps.navigation.getParam('city')
   const targetPath: string = ownProps.navigation.getParam('path') || `/${targetCity}/${language}`
 
-  const cities = state.cities.json
+  const cities = state.cities
 
-  if (!cities) {
+  if (!cities.json || cities.error) {
```
maxammann: A check for `!cities` usually doesn't happed, but can happen when the state gets rehydrated in a wrong way.

maxammann: Change to `  if (!cities || !cities.json || cities.error) {`

src/modules/common/hocs/withError.js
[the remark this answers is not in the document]
maxammann: Hm I can not reproduce that flow error

src/modules/common/components/Failure.js
```diff
@@ -0,0 +1,34 @@
...
+  t: TFunction
+|}
+
+export class Failure extends React.Component<PropsType> {
+  render () {
+    const {t} = this.props
+    return <ViewContainer>
+      <Icon name='frown-o' size={50} />
+      <Text>
+        {t('loadingFailed')}
```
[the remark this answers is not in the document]
maxammann: If the error appears in the console we're fine.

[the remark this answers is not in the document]
NKlug: I'm logging it to the console now.

src/modules/error/components/Failure.js
```diff
@@ -0,0 +1,36 @@
...
+type PropsType = {|
+  error: Error,
+  t: TFunction
+|}
+
+export class Failure extends React.Component<PropsType> {
+  render () {
+    const {error, t} = this.props
+
+    console.error(error)
```
klinzo: Should we log this error here?

NKlug: Is there a better place to log it? I think this is just the most straightforward.

maxammann: Console logs should be avoided in production mode: https://facebook.github.io/react-native/docs/performance#using-consolelog-statements

Also test would output the errors. Maybe there is a better place for logging. e.g. the place where the exception was thrown.

src/modules/error/components/Failure.js
```diff
@@ -0,0 +1,35 @@
...
+  error: Error,
+  t: TFunction
+|}
+
+export class Failure extends React.Component<PropsType> {
+  render () {
+    const { t } = this.props
+
+    return <ViewContainer>
+      <Icon name='frown-o' size={50} />
```
maxammann: I think I only used the material icons so far. You can use this icon instead: https://material.io/tools/icons/?icon=sentiment_very_dissatisfied&style=baseline

PR #5, opened 2018-11-27
NATIVE-29: Upgrade packages

maxammann:


android/app/src/main/java/com/integreat/MainApplication.java
```diff
@@ -3,9 +3,10 @@
 import android.app.Application;
 
 import com.facebook.react.ReactApplication;
+import com.swmansion.gesturehandler.react.RNGestureHandlerPackage;
```
NKlug: For what do we need the GestureHandler atm?

maxammann: This is a new dependency of some package. not sure which though.

PR #8, opened 2018-12-04
NATIVE-30: Upgrade react-native-view-pdf

maxammann:


thimmy687:
CI Server seems to be out of memory...

Please fix

PR #9, opened 2018-12-11
NATIVE-19: Add icons and splash screen

maxammann:


android/.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,29 @@
+<component name="ProjectCodeStyleConfiguration">
```
NKlug: Why do we need another .idea folder in the android folder?

android/app/src/main/res/values/strings.xml
```diff
@@ -1,3 +1,3 @@
 <resources>
-    <string name="app_name">Integreat</string>
+    <string name="app_name">Integreat (RN)</string>
```
NKlug: Why RN?

android/.idea/codeStyles/Project.xml
maxammann: That way you can open the android project in Android Studio/Intellij and debug it from there.

android/app/src/main/res/values/strings.xml
maxammann: So we can distinguish it from the current app.

android/.idea/codeStyles/Project.xml
thimmy687: I also like this approach

PR #11, opened 2018-12-27
NATIVE-60: Copy extras from webapp

NKlug:
I misnamed some commits :(


src/modules/common/components/Tile.js
```diff
@@ -43,12 +50,7 @@ class Tile extends React.Component<PropsType> {
   }
 
   getTile (): React.Node {
-    const tile = this.props.tile
-    if (!tile.isExternalUrl) {
-      return <TouchableOpacity onPress={this.onTilePress}>{this.getTileContent()}</TouchableOpacity>
-    }
-
-    return null
+    return <TouchableOpacity onPress={this.onTilePress}>{this.getTileContent()}</TouchableOpacity>
```
klinzo: You could just inline this, no need for a getTile method anymore.

src/modules/extras/ExtrasConfig.js
```diff
@@ -0,0 +1,13 @@
...
+import Hashids from 'hashids'
+import { WohnenOfferModel } from '@integreat-app/integreat-api-client'
+
+export const SPRUNGBRETT_ROUTE = 'SprungbrettExtra'
+export const SPRUNGBRETT_EXTRA = 'sprungbrett'
+
+export const WOHNEN_ROUTE = 'WohnenExtra'
+export const WOHNEN_EXTRA = 'wohnen'
+
+export const hash = (offer: WohnenOfferModel) =>
```
klinzo: Maybe rename this to `hashWohnenOfferModel` or sth like that.

src/modules/extras/components/Extras.js
```diff
@@ -0,0 +1,57 @@
...
+} from '../ExtrasConfig'
+import { ScrollView } from 'react-native'
+
+type PropsType = {|
+  extras: Array<ExtraModel>,
+  navigateToExtras: (path: string, isExternalUrl: boolean) => void,
+  t: TFunction
+|}
+
+export default class Extras extends React.Component<PropsType> {
```
klinzo: Rename to `ExtraTiles` or sth like that. Or is this a convention you agreed on?

src/modules/extras/components/Extras.js
```diff
@@ -0,0 +1,57 @@
...
+import {
+  SPRUNGBRETT_EXTRA,
+  SPRUNGBRETT_ROUTE,
+  WOHNEN_EXTRA, WOHNEN_ROUTE
+} from '../ExtrasConfig'
+import { ScrollView } from 'react-native'
+
+type PropsType = {|
+  extras: Array<ExtraModel>,
+  navigateToExtras: (path: string, isExternalUrl: boolean) => void,
```
klinzo: Shouldn't this be singular (`navigateToExtra`)?

src/modules/extras/containers/ExtrasContainer.js
```diff
@@ -0,0 +1,89 @@
...
+  const map = new Map()
+  Object.keys(jsonPost).forEach(key => map.set(key, jsonPost[key]))
+  return map
+}
+
+const mapStateToProps = (state: StateType, ownProps) => {
+  const language: string = state.language
+
+  const targetCity: CityModel = ownProps.navigation.getParam('cityModel')
+  console.log(targetCity)
```
klinzo: Remove this.

src/modules/extras/containers/ExtrasContainer.js
```diff
@@ -0,0 +1,89 @@
...
+    alias: extra.alias,
+    title: extra.name,
+    path: extra.url,
+    // Important. Disables caching of the extras' thumbnails. We want to do it manually at some point.
+    thumbnail: extra.thumbnail,
+    postData: extra.post ? createPostMap(extra.post) : null
+  }))
+
+  const navigateToExtras = (path: string, isExternalUrl: boolean, offerHash: ?string = null) => {
+    console.log(path)
```
klinzo: Remove console.log and rename to `navigateToExtra`

src/modules/wohnen/components/OfferDetail.js
```diff
@@ -0,0 +1,165 @@
...
+`
+
+const ListElement = styled.View`
+  border-bottom-width: 2px;
+  border-bottom-color: ${props => props.theme.colors.themeColor};
+  margin-bottom: 10px;
+`
+
+class OfferDetail extends React.PureComponent<PropsType> {
+  translate (type: 'runningServices' | 'additionalServices' | 'rooms', keys: Array<string>): Array<string> {
```
klinzo: Why do we need to translate here? At least in the WebApp we agreed to only provide the Wohnraumbörse Extra in German. This looks quite strange anyway.

src/modules/wohnen/components/OfferListItem.js
```diff
@@ -0,0 +1,48 @@
...
+  offer: WohnenOfferModel,
+  navigateToOffer: () => void
+|}
+
+class OfferListItem extends React.PureComponent<PropsType> {
+  render () {
+    const {offer, navigateToOffer} = this.props
+
+    if (offer.formData instanceof WohnenFormData) {
+      const specificOffer: WohnenOfferModel = offer
```
klinzo: Couldn't we inline this variable?

src/modules/app/components/Navigator.js
```diff
@@ -29,10 +33,25 @@ const createNavigationScreen = (component, header = null) => {
...
+export const ExtrasStack = createStackNavigator(
+  {
+    'Extras': ExtrasContainer,
+    [WOHNEN_ROUTE]: WohnenExtraContainer,
+    [SPRUNGBRETT_ROUTE]: SprungbrettExtraContainer
+  },
+  {
+    initialRouteName: 'Extras',
+    defaultNavigationOptions: {
+      header: null
```
maxammann: Which header are use using the the extras?

src/modules/extras/ExtrasConfig.js
maxammann: We should not need the hashids and hash function. This is only needed in the web because an offer does not have an ID.

src/modules/wohnen/components/OfferDetail.js
maxammann: I think we use the keys in the webapp aswell. We only translate the keys to German though.

src/modules/extras/ExtrasConfig.js
```diff
@@ -0,0 +1,13 @@
+// @flow
```
maxammann: This file is quite weird. It has a name like it is a class but it isn't.
If we remove the hash function we can just put the other exports (which are only string constants) to the routing.

src/modules/extras/components/Extras.js
NKlug: We also have `Categories`, so I would just leave it like that.

PR #12, opened 2019-01-22
Upgrade packages

NKlug:
Please test the app before approving. On my phone the redirect after a click on a tile feels quite slow, slower than before, but I'm not sure if that has something to do with the upgrades. So please test this on your device, too.


android/app/build.gradle
```diff
@@ -150,6 +150,14 @@ android {
 }
 
 dependencies {
+    compile project(':rn-fetch-blob')
```
maxammann: These are duplicates as they are already included below.

android/settings.gradle
```diff
@@ -1,8 +1,8 @@
 rootProject.name = 'Integreat'
+include ':react-native-screens'
+project(':react-native-screens').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-screens/android')
 include ':react-native-gesture-handler'
 project(':react-native-gesture-handler').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-gesture-handler/android')
-include ':react-native-view-pdf'
```
maxammann: This change is also odd. why is the pdf view removed?

ios/Podfile
```diff
@@ -52,6 +52,8 @@ target 'Integreat' do
   pod 'RNVectorIcons', :path => '../node_modules/react-native-vector-icons'
   pod 'RNGestureHandler', :path => '../node_modules/react-native-gesture-handler/ios'
 
+  pod 'RNScreens', :path => '../node_modules/react-native-screens'
```
maxammann: This needs to be tested before merge.

maxammann: Also `pod install` should be run to update the pod lock file.

android/settings.gradle
NKlug: It was in there twice, so I removed one occurence.

android/app/build.gradle
```diff
@@ -150,10 +150,12 @@ android {
 }
 
 dependencies {
+    compile project(':@mapbox_react-native-mapbox-gl')
```
maxammann: This is still a duplicate

ios/Integreat.xcodeproj/project.pbxproj
```diff
@@ -27,6 +26,21 @@
...
 		ADBDB9381DFEBF1600ED6528 /* libRCTBlob.a in Frameworks */ = {isa = PBXBuildFile; fileRef = ADBDB9271DFEBF0700ED6528 /* libRCTBlob.a */; };
+		50A86AA641004ED5A039EA7F /* AntDesign.ttf in Resources */ = {isa = PBXBuildFile; fileRef = E3105738B7C34990835457F8 /* AntDesign.ttf */; };
+		BB16A74067834BB88022B535 /* Entypo.ttf in Resources */ = {isa = PBXBuildFile; fileRef = BB51B58C720B47A69EF40EA8 /* Entypo.ttf */; };
+		6556375C270B4B3282F952C8 /* EvilIcons.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 93E9F91A034743CDA0862F4D /* EvilIcons.ttf */; };
+		DCDE89D72B3F445BBE516AD7 /* Feather.ttf in Resources */ = {isa = PBXBuildFile; fileRef = D772DA281D684A94956032D4 /* Feather.ttf */; };
+		706B89ABE414487E91E82687 /* FontAwesome.ttf in Resources */ = {isa = PBXBuildFile; fileRef = D1C82620A5C64C63AE8FDF99 /* FontAwesome.ttf */; };
+		9A37E2C4B41D486485ADD9CD /* FontAwesome5_Brands.ttf in Resources */ = {isa = PBXBuildFile; fileRef = AD5F005B99D345608B0690D1 /* FontAwesome5_Brands.ttf */; };
+		DDA090C57F1B407780220F4B /* FontAwesome5_Regular.ttf in Resources */ = {isa = PBXBuildFile; fileRef = EE808817D5AC4437A4774120 /* FontAwesome5_Regular.ttf */; };
```
maxammann: Somehow here are still ttfs files. Which script did add these?

ios/Integreat/Info.plist
```diff
@@ -49,9 +49,8 @@
 		<string>Octicons.ttf</string>
 		<string>SimpleLineIcons.ttf</string>
 		<string>Zocial.ttf</string>
+		<string>AntDesign.ttf</string>
```
maxammann: Same as here and a few other places.

NKlug:
@maxammann I know hopefully removed all ttf files/references that were added. Maybe we should at some point look for the reason they are added by rn link in the first place.

PR #15, opened 2019-02-08
NATIVE-59: Copy events

NKlug:
Caption, date and location are currently not working when selecting an event. This will be taken care of in NATIVE-84.


src/modules/common/components/ListItem.js
```diff
@@ -13,52 +19,42 @@ const StyledTouchableOpacity = styled.TouchableOpacity`
...
+  padding: 0 10px;
 `
 
 const Title = styled.Text`
   font-weight: 700;
 `
 
 type PropsType = {|
-  thumbnail?: string,
+  thumbnail?: {uri: string} | string,
```
maxammann: This can be renamed to `thumbnailUrl` with the type `string`

src/modules/common/components/Page.js
```diff
@@ -54,13 +57,16 @@ class Page extends React.Component<PropType> {
   }
 
   render () {
+    const {title, children, theme, content, files} = this.props
     return (
       <>
-        <WebContainer theme={this.props.theme}>
+        <Caption title={title} />
+        {children}
```
maxammann: Children can have arbitrary heights. So this will not work with together with the webview.  Maybe a refactor to `ContentView` which renders the webview and a component for each different usage  of `ContentView`  would make sense.

So you would have a EventPage and CategoriesPage which both render a `ContentView` ContentView would tage a required height.

src/modules/common/components/ListItem.js
NKlug: It can be renamed, but the type has to stay like it is. We sometimes want to display a specific thumbnail from a link (hence `{uri: string}`) but if none is given display a Placeholder thumbnail from assets, which we have to pass as a string. See [here](https://github.com/Integreat/integreat-react-native-app/pull/15/files#diff-5f5945d35181867cf92ff917f4eb2e98R33).

src/modules/common/components/Page.js
NKlug: How exactly does the webview render? I don't quite understand why it shouldn't just move down automatically, I mean everything is in a ScrollView and has a 'column' flex direction.  
What do you mean by passing a height to the content view? How would we know how much space a specific component needs on each device?

maxammann: Each component next to a WebView needs to have a static height. The WebView brings its own scrolling behaviour. Mixing it with a ScrollView does not work afaik.

maxammann: The height of the device specific statusbar can be calculated.

src/modules/common/components/ListItem.js
maxammann: If would be interresting to know what value EventPlaceholder3 has here: import EventPlaceholder3 from '../assets/EventPlaceholder3.jpg'

Its probably already an URI or URL. So this type can probably be simplified.
Or use the type of the Image if you can find it somewhere: ImageSourcePropType https://facebook.github.io/react-native/docs/image#source

maxammann:
@klinzo @Simple-codinger ping

src/modules/common/components/Page.js
maxammann: I think only this needs to changed here. Else this is ready for merge.

You can also create a separate task to figure out a way to handle the webview.

src/modules/events/components/EventListItem.js
```diff
@@ -0,0 +1,47 @@
...
+import * as React from 'react'
+import { View } from 'react-native'
+import styled from 'styled-components'
+import { EventModel } from '@integreat-app/integreat-api-client'
+import ListItem from '../../../modules/common/components/ListItem'
+import EventPlaceholder1 from '../assets/EventPlaceholder1.jpg'
+import EventPlaceholder2 from '../assets/EventPlaceholder2.jpg'
+import EventPlaceholder3 from '../assets/EventPlaceholder3.jpg'
+
+const WrappedText = styled.Text`
```
klinzo: Can't this be removed if you don't style anything here?

src/modules/common/components/ListItem.js
NKlug: EventPlaceholder has type `number`. As far as I can see there is no other way to do it.

src/modules/common/components/Page.js
NKlug: I'll have a look at it in NATIVE-32.

src/modules/events/components/EventListItem.js
NKlug: Removed it.

PR #17, opened 2019-03-15
NATIVE-32: Fix Webview height

NKlug:
Fixes the height of the webview (also upgrades the webview version).

### DIFF
--- a/src/modules/common/renderHtml.js
+++ b/src/modules/common/renderHtml.js
@@ -40,8 +40,7 @@ const renderJS = (files: FilesStateType) => `
   function adjustHeight() {
     if (window.postMessage.length !== 1){
       setTimeout(adjustHeight, 200);
-    }
-    else {
+    } else {
       container.setAttribute('style', 'padding: 1px 0;'); // Used for measuring collapsed vertical margins
       window.ReactNativeWebView.postMessage(container.getBoundingClientRect().height - 2);
       container.setAttribute('style', '');
--- a/src/modules/common/renderHtml.js
+++ b/src/modules/common/renderHtml.js
@@ -35,6 +35,17 @@ const renderJS = (files: FilesStateType) => `
   
   document.body.style.display = 'block'
 })();
+(function waitForBridge() {
+    if (window.postMessage.length !== 1){
+      setTimeout(waitForBridge, 200);
+    }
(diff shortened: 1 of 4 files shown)


src/modules/common/renderHtml.js
```diff
@@ -32,18 +33,91 @@ const renderJS = (files: FilesStateType) => `
...
   }
+})();
+
+(function() {
+  var container = document.getElementById('measure-container')
+  function adjustHeight() {
+    if (window.postMessage.length !== 1){
+      setTimeout(adjustHeight, 200);
+    }
+    else {
```
maxammann: Formatting is a bit off here

src/modules/common/renderHtml.js
```diff
@@ -32,18 +33,91 @@ const renderJS = (files: FilesStateType) => `
...
+  <meta name="viewport" content="initial-scale=1.0, maximum-scale=1.0">
+  <style>
+    html, body {
+        margin: 0;
+        padding: 0;
+        
+        /*font-family: \${theme.fonts.contentFontFamily};*/   
+        font-size: ${theme.fonts.contentFontSize};
+        line-height: ${theme.fonts.contentLineHeight};
+        /*\${props => props.centered && css\`
```
maxammann: Can probably be removed.

src/modules/common/renderHtml.js
```diff
@@ -32,18 +33,91 @@ const renderJS = (files: FilesStateType) => `
       item.src = '${URL_PREFIX}' + newSrc
     }
   }
+})();
+
+(function() {
+  var container = document.getElementById('measure-container')
```
maxammann: Can we use the body here?

src/modules/common/renderHtml.js
Schedulaar: no probably not; we'll need that once we use webview for categorylists as well

src/modules/common/renderHtml.js
Schedulaar: no, for some reason, body seems to behave indeterminately (at least for me)

src/modules/common/renderHtml.js
NKlug: Done.

PR #19, opened 2019-03-16
NATIVE-6: Add share button

a-kud:
As a user I'd like to able to share app link.

Fixes [NATIVE-6](https://issues.integreat-app.de/browse/NATIVE-6)
Screenshot_1557951610
Screenshot_1557951817
image


src/modules/layout/components/Header.js
```diff
@@ -123,6 +124,24 @@ class Header extends React.PureComponent<PropsType, StateType> {
     this.getNavigation().navigate('ChangeLanguageModal')
   }
 
+  onShare = async () => {
+    const url = 'https://integreat-app.de/'
+    const message: string = Platform.select({
```
NKlug: Is there a reason why we can't display the same message on both platforms?

src/modules/layout/components/Header.js
```diff
@@ -123,6 +124,24 @@ class Header extends React.PureComponent<PropsType, StateType> {
     this.getNavigation().navigate('ChangeLanguageModal')
   }
 
+  onShare = async () => {
+    const url = 'https://integreat-app.de/'
```
NKlug: I think it would be nice to get a link to the current page here, so if I'm on route 'testumgebung/en/welcome/' I want to get a link to 'https://integreat.app/testumgebung/en/welcome/'.  The path (if navigated to Categories) should be available in the store.
Also use our new domain 'https://integreat.app'.

maxammann: Yes we need to link to the current selected page. You probably need to merge in develop for that.

src/modules/layout/components/Header.js
a-kud: I read this thread https://expo.canny.io/feature-requests/p/share-api-support-for-sharing-urls-on-android
but I didn't have a chance to test it on iOS

maxammann: Alright I think this is the way to go on ios vs android.

Only i18n is missing for the texts. You can add a constant to this file: https://docs.google.com/spreadsheets/d/1kMxVlqzwjjDoARO5Ss5oMNUEjqubT0Ps8NB7cguP-yU/

Then just add a key to the translation to locales.json
Also create a task to add the translations to the project once the Konstanten sheet has been translated.

maxammann: This is still relevant.

src/locales.json
```diff
@@ -354,5 +354,11 @@
       "feedbackSent": "Feedback sent",
       "close": "Close"
     }
+  },
+  "heading": {
```
maxammann: I think "header" would be better because heading is more like a title

src/modules/layout/containers/HeaderContainer.js
```diff
@@ -1,16 +1,25 @@
...
-
 import Header from '../components/Header'
+// import { connect } from 'react-redux'
+// import { withTheme } from 'styled-components'
+// import { CategoryModel } from '@integreat-app/integreat-api-client'
+
+// const mapStateToProps = ({ categories }: CategoryModel) => categories
+
+// // $FlowFixMe
+// export default connect(mapStateToProps)(withTheme(Header))
```
klinzo: Why is this commented out?

src/modules/layout/containers/HeaderContainer.js
```diff
@@ -1,16 +1,25 @@
...
 import { connect } from 'react-redux'
 import React from 'react'
 import { withNavigation } from 'react-navigation'
+import { translate } from 'react-i18next'
 import { availableLanguagesSelector } from '../../common/selectors/availableLanguagesSelector'
 
 const mapStateToProps = (state: StateType, ownProps) => ({
-  availableLanguages: availableLanguagesSelector(state, ownProps)
+  availableLanguages: availableLanguagesSelector(state, ownProps),
+  routeMapping: state.cityContent.categoriesRouteMapping
```
klinzo: This way, it will just work for sharing categories.

maxammann: I created a task for this so we can merge this one: https://issues.integreat-app.de/browse/NATIVE-224

maxammann:
@a-kud feel free to merge this.

Schedulaar:
You might want to have a look at https://github.com/Integreat/integreat-react-native-app/pull/19/commits/b82cd2353f0b66ecc6574dadd58af060fb5db086 again; otherwise I will merge this tonight...

PR #22, opened 2019-03-19
NATIVE-104: Fix push reducer by adding the children to the models

maxammann:


src/modules/endpoint/reducers/pushCategory.js
```diff
@@ -36,6 +37,7 @@ const pushCategory = (
 
   forEachTreeNode(root, node => categoriesMap.getChildren(node), depth, (node, children) => {
     resultModels[node.path] = node
+    Object.assign(resultModels, keyBy(children, child => child.path))
```
maxammann: I'm a bit unsure why we need to add the children. I mean they should be added by `resultModels[node.path] = node`

NKlug: I agree. I think all this line does is add 1 to the `depth`. This seems a bit odd to me.

src/modules/endpoint/reducers/pushCategory.js
```diff
@@ -3,6 +3,7 @@
 import type { CategoriesStateType } from '../../app/StateType'
 import type { PushCategoryActionType } from '../../app/StoreActionType'
 import { CategoryModel } from '@integreat-app/integreat-api-client'
+import { keyBy } from 'lodash/collection'
 
 const forEachTreeNode = (
```
Schedulaar:
```js
const forEachTreeNode = (
  root: CategoryModel,
  resolveChildren: CategoryModel => Array<CategoryModel>,
  depth: number,
  nodeAction: (CategoryModel, Array<CategoryModel>) => void
) => {
  if (depth == 0) {
    nodeAction(root, null)
  } else {
    const children = resolveChildren(root)
    nodeAction(root, children)
    children.forEach(child => forEachTreeNode(child, resolveChildren, depth - 1, nodeAction))
  }
}
```

src/modules/endpoint/reducers/pushCategory.js
```diff
@@ -36,6 +37,7 @@ const pushCategory = (
 
   forEachTreeNode(root, node => categoriesMap.getChildren(node), depth, (node, children) => {
     resultModels[node.path] = node
+    Object.assign(resultModels, keyBy(children, child => child.path))
     resultChildren[node.path] = children.map(child => child.path)
   })
```
Schedulaar:
```js
forEachTreeNode(root, node => categoriesMap.getChildren(node), depth, (node, children) => {
    resultModels[node.path] = node
    if (children !== null) {
      resultChildren[node.path] = children.map(child => child.path)
    }
  })
```

src/modules/endpoint/reducers/pushCategory.js
```diff
@@ -4,15 +4,27 @@ import type { CategoriesStateType } from '../../app/StateType'
 import type { PushCategoryActionType } from '../../app/StoreActionType'
 import { CategoryModel } from '@integreat-app/integreat-api-client'
 
+/**
+ * Iterates through a tree until depth is reached. A depth of 0 means only the root is visited.
+ * A depth of 1 means the root and the children of it are visited. Please note that for each children of the root
+ * nodeAction is called once with the category and null for the children as parameters.
```
Schedulaar: Please note that for each node with depth `depth` null is passed as argument for `children` in nodeAction.

PR #23, opened 2019-03-19
NATIVE-79: Setup testing

NKlug:


src/modules/common/components/__tests__/PageDetail.spec.js
```diff
@@ -0,0 +1,15 @@
...
+import { render } from 'react-native-testing-library'
+import React from 'react'
+import PageDetail, { Identifier } from '../PageDetail'
+
+describe('PageDetail', () => {
+  it('should display the given identifier followed by a colon', () => {
+    const { getByType } = render(
+      <PageDetail identifier={'Test Identifier'} information={'Some important information'} />
+    )
+    const identifier = getByType(Identifier).props.children.join('')
```
NKlug: I added this test just to try and demonstrate how render works here. For example getting the text of a component feels somewhat more complex than enzymes `.text()`, but I guess we have to settle with it.
Should I remove this test again before merging?

maxammann: Hm I don't really get why we call `children` on `props`.

maxammann: Probably should be removed again yeah. Or you also assert that the information prop is missing

NKlug: Maybe just calling `.children` also works, [this](https://reactjs.org/docs/test-renderer.html#testinstance) is the datatype usually returned by the library.

Schedulaar: I don't see a problem with this kind of access to the text output. We've never used .text() with enzyme in the webapp afaik. In this kind of scenario we would probably use a snapshot...

PR #27, opened 2019-03-28
NATIVE-140: Fix translation for events

maxammann:


src/modules/endpoint/reducers/cityContentReducer.js
```diff
@@ -15,8 +15,8 @@ export default (
       return pushCategory(state, action)
     case 'PUSH_EVENT':
       return pushEvent(state, action)
-    case 'SWITCH_CITY_CONTENT_LANGUAGE': // todo:  Not working
-      return switchLanguage(state, action)
+    case 'MORPH_CONTENT_LANGUAGE':
```
Schedulaar: morph? 🤨 I like "switch" better. Even "change" sounds better to me...

src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -34,8 +34,8 @@ export default function * loadCityContent (database: MemoryDatabase, newCity: st
 
   database.addCacheEntries(resourceCache)
 
-  //yield call(() => database.writeCategories())
-  //yield call(() => database.writeResourceCache())
-  //yield call(() => database.readCategories())
-  //yield call(() => database.readResourceCache())
+  // yield call(() => database.writeCategories())
```
Schedulaar: Add comment or remove

src/routes/events/components/Events.js
```diff
@@ -59,7 +59,7 @@ export default class Events extends React.Component<PropsType> {
           </Page>
         </ScrollView>
       }
-      const error = new ContentNotFoundError({type: 'event', id: event.id, city, language})
+      const error = new ContentNotFoundError({type: 'event', id: path, city, language})
```
Schedulaar: Maybe ContentNotFoundError should be adjusted as well (id -> path)

src/modules/endpoint/reducers/cityContentReducer.js
maxammann: I tried to use something extraordinary. `SWITCH_` is already used. See the changes for that.
Change is too generic in my opinion. Also the term Morph is used in Computing: https://en.wikipedia.org/wiki/Morph#Technology

We have a problem because we need 2 terms for kind of the same thing: Like "PUSH" and "FETCH" there is "MORPH" and "SWITCH"

src/routes/events/components/Events.js
maxammann: The id has type `numeric | string`. Therefore it was never indented to be the numeric id we previously used. I think it is fine not to rename it here.

PR #28, opened 2019-03-28
NATIVE-138: Create and update android app icons

thimmy687:


maxammann:
Did you kind of follow this way to generate these icons?


- Open your project in Android Studio
- Right click on the app folder, then click on New and then click on Image Asset
- In the Icon Type field select Launcher Icons (Adaptative and Legacy)
- In the pathfield, select your image file
- Once done, click on the Nextbutton and then on the finish button

We used this guide to generate the icons initially: https://github.com/bamlab/generator-rn-toolbox/blob/master/generators/assets/README.md#assets-setup

We should try to stay compatible to this generator. Also is the source for the new icons the `assets/` driectory in the root of the repo?

dkehne:
Thanks @thimmy687 for already delivering it! :)  
backup:
[Uploading adaptive-icons.zip…]()

android/app/src/main/res/drawable/ic_launcher_background.xml
```diff
@@ -0,0 +1,74 @@
+<?xml version="1.0" encoding="utf-8"?>
```
maxammann: What is this file about? This is kind of a green background. Do we need this?
I think as icons need to be opaque we have to decide for a possible background color: https://stackoverflow.com/questions/47722958/transparent-adaptive-icon-background

thimmy687: Hm.. I used the Android studio to generate the icons... 

I expected that all created files have a sense, but I'm not sad if we remove this if it is not necessary

maxammann: I think this is required but should be black or white. Depending on what looks better.

dkehne: If you create a new standard project in Android studio it's taking a default color. Maybe the generator did not work properly here. Every ic_launcher_background color can just set to #FBDA16 - we did this also with other #33FFFF-appearances.

thimmy687: Ok, can someone help to change this?

I guess I will not be able to do this in the next week's :/

maxammann: As this is in the Release Epic it is not important to finish early.

maxammann:
Can be reopened if this is useful.

PR #30, opened 2019-04-02
NATIVE-62: Fix webview width

NKlug:


src/modules/common/components/Page.js
```diff
@@ -120,6 +133,7 @@ class Page extends React.Component<PropType, StateType> {
 
             onMessage={this.onMessage}
             style={{height: height}}
+            width={width}
```
Schedulaar: try to use the style attribute here for consistency

src/modules/common/components/Page.js
```diff
@@ -46,15 +47,26 @@ type PropType = {
 }
 
 class Page extends React.Component<PropType, StateType> {
+  // noinspection JSDuplicatedDeclaration
   onMessage: (event: WebViewMessageEvent) => void
+  // noinspection JSDuplicatedDeclaration
```
Schedulaar: try to avoid these with arrow-functions.

src/modules/common/components/Page.js
NKlug: Done.

src/modules/common/components/Page.js
NKlug: Done.

PR #31, opened 2019-04-02
NATIVE-62: Dashboard buttons

NKlug:
This is a first draft of the new Dashboard buttons. We will use this until Anna-Licia finishes the new "new Design".

Known Problems:
* Long text overflows
* "Orte" does nothing
* Captions are not translated yet


src/modules/common/models/TileModel.js
```diff
@@ -47,4 +53,12 @@ export default class TileModel {
   get postData (): ?Map<string, string> {
     return this._postData
   }
+
+  get onTilePress (): ?() => void {
```
NKlug: I just extended the existing `TileModel`s, do you think it's better to create individual Models?

src/modules/common/components/NavigationTile.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  getTileContent (): React.Node {
+    const {tile, theme} = this.props
+    const imageSource = typeof tile.thumbnail === 'number' ? tile.thumbnail : {
+      uri: tile.thumbnail,
+      priority: FastImage.priority.normal,
+      // disable caching, we want to do it manually
+      headers: {'Cache-Control': 'no-cache, no-store, must-revalidate'},
+      cache: FastImage.cacheControl.web
```
maxammann: It seems like this is duplicate code in the. Not sure whether it makes sense to de-duplicate this

src/modules/common/components/NavigationTile.js
```diff
@@ -0,0 +1,98 @@
...
+  border-radius: ${NEWS_DOT_RADIUS / 2};
+  background-color: #EE5353;
+  color: #FFFFFF;
+  elevation: 5;
+`
+
+/**
+ * Displays a single Tile
+ */
+class Tile extends React.Component<PropsType> {
```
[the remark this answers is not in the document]
NKlug: Done.

src/modules/common/components/NavigationTile.js
NKlug: Done.

PR #34, opened 2019-04-03
NATIVE-135: Add mapper for events

maxammann:


src/modules/endpoint/MemoryDatabase.js
```diff
@@ -29,6 +29,29 @@ type ContentCategoryJsonType = {|
...
+    end_date: string,
+    all_day: boolean
+  },
+  location: {
+    address: string,
+    town: string,
+    postcode: ?string,
+    latitude: ?string,
+    longitude: ?string
+  }
```
Schedulaar: should we make this an exact object type here (also above)?

src/modules/endpoint/MemoryDatabase.js
```diff
@@ -184,6 +206,72 @@ class MemoryDatabase {
     }))
   }
 
+  writeEvents (): Promise<number> {
```
Schedulaar: why does writeEvents return a Promise, but readEvents is async? Can we make writeEvents an async function?

NKlug: I fixed this in #35, you don't necessarily have to fix it here.

src/modules/endpoint/MemoryDatabase.js
maxammann: Hm I kind of want that this is the default behaviour, but yes yes :D

PR #35, opened 2019-04-05
NATIVE-92: Loading and storing

NKlug:


src/modules/endpoint/sagas/loadLanguages.js
```diff
@@ -5,9 +5,17 @@ import { call } from 'redux-saga/effects'
...
 import request from '../request'
 import { baseUrl } from '../constants'
+import MemoryDatabase from '../MemoryDatabase'
 
-export default function * loadLanguages (city: string): Saga<?Array<LanguageModel>> {
+export default function * loadLanguages (city: string, database: MemoryDatabase): Saga<void> {
+  yield call(database.readLanguages)
+
+  if (database.languagesLoaded()) {
+    console.dir(database.languages)
```
Schedulaar: remove  console.dir

src/modules/endpoint/MemoryDatabase.js
```diff
@@ -130,12 +136,17 @@ class MemoryDatabase {
...
+  eventsLoaded = () => this._events !== null
+
   /**
    * @returns {Promise<void>} which resolves to the number of bytes written or rejects
    */
-  writeCategories (): Promise<number> {
+  writeCategories = async () => {
     if (!this.categoriesMap) {
-      throw new Error('MemoryDatabase does not have data to save!')
+      console.warn('MemoryDatabase does not have data to save!')
```
Schedulaar: why did you make these errors warnings?

src/modules/endpoint/MemoryDatabase.js
```diff
@@ -184,29 +195,73 @@ class MemoryDatabase {
...
+  writeEvents = async () => {
+    if (!this._events) {
+      console.warn('MemoryDatabase does not have data to save!')
+      return
+    }
+    const path = this.getContentPath('events')
+    await this.writeFile(path, JSON.stringify(this._events))
+  }
+
+  readResourceCache = async () => {
```
Schedulaar: we should not use lambda functions as function declarations if we don't need them.
Instead do sth like
```js
async readResourceCache () {
```

src/modules/endpoint/MemoryDatabase.js
NKlug: Not really sure. Reverting it.

src/modules/endpoint/sagas/loadLanguages.js
NKlug: Done.

src/modules/endpoint/MemoryDatabase.js
NKlug: Done.

src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -28,25 +28,29 @@ const createErrorMessage = (fetchResult: FetchResultType) => {
   }, '')
 }
 
-export default function * fetchResourceCache (city: string, language: string, fetchMap: FetchMapType): Saga<ResourceCacheStateType> {
+export default function * fetchResourceCache (city: string, language: string, fetchMap: FetchMapType, database: MemoryDatabase): Saga<void> {
+  yield call(database.readResourceCache)
+
+  if (isEmpty(fetchMap)) {
+    return
```
maxammann: I think we should always fire an action here. In this case a success action should be fired.

NKlug: Done.

PR #38, opened 2019-04-15
NATIVE-165: Failed resources downloads

Schedulaar:


NKlug:
What does currently happen when one navigates to a page that contains a resource whose download has failed? I'm not sure if we want to display an error message or just nothing there. But anyways it should not crash :)

src/modules/categories/components/Categories.js
```diff
@@ -97,7 +97,7 @@ class Categories extends React.Component<PropsType> {
 
     if (children.length === 0) {
       // last level, our category is a simple page
-      const files = this.getLocalResourceCache(category)
+      const files = this.getFileCache(category)
```
maxammann: This should be named ResourceCache. The term FileCache should no longer exist in the app.
Also "local" is quite correct because it is not the global resource cache of the app.

src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -53,27 +49,31 @@ export default function * fetchResourceCache (city: string, language: string, fe
     const failureResults = pickBy(results, result => !!result.errorMessage)
     if (!isEmpty(failureResults)) {
       const message = createErrorMessage(failureResults)
-
       const failed: ResourcesFetchFailedActionType = {type: `RESOURCES_FETCH_FAILED`, city, language, message}
+      console.warn(message)
+      // todo: we might remember which files have failed to retry later (internet connection of client could have failed)
```
maxammann: Feel free to create a task for that todo.

src/modules/categories/components/Categories.js
Schedulaar: the type is called FileCache.

src/modules/endpoint/sagas/fetchResourceCache.js
Schedulaar: done

src/modules/categories/components/Categories.js
maxammann: Oh oke you are right.. too many file name changes :sweat_smile:

PR #40, opened 2019-04-17
NATIVE-141: Better database interface

NKlug:
* MemoryDatabase is now called DataContainer
* Loading and Storing has been moved to a DatabaseConnector
* Loading and Storing is handled solely by the DataContainer
* The DataContainer and the "database" are always consitent (and synchronized)
* All getters and setters in the DataContainer now return promises


src/modules/endpoint/DataContainer.js
```diff
@@ -0,0 +1,250 @@
...
+  setResourceCache = async (resourceCache: ResourceCacheStateType) => {
+    if (this._context === null) {
+      throw Error('Context has not been set yet.')
+    }
+    await this._databaseConnector.storeResourceCache(resourceCache, this._context)
+    this._resourceCache = resourceCache
+  }
+
+  categoriesMapLoaded (): boolean {
+    return this._categoriesMap === null
```
Schedulaar:
```suggestion
    return this._categoriesMap !== null
```

src/modules/endpoint/DataContainer.js
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  languagesLoaded (): boolean {
+    return this._languages === null
```
Schedulaar:
```suggestion
    return this._languages !== null
```

src/modules/endpoint/DataContainer.js
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  eventsLoaded (): boolean {
+    return this._events === null
```
Schedulaar:
```suggestion
    return this._events !== null
```

src/modules/endpoint/DataContainer.js
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  resourceCacheLoaded (): boolean {
+    return this._resourceCache === null
```
Schedulaar:
```suggestion
    return this._resourceCache !== null
```

src/modules/endpoint/DataContainer.js
```diff
@@ -0,0 +1,250 @@
...
+
+  /**
+   * Returns whether the ResourceCache have been loaded or not.
+   */
+  resourceCacheLoaded (): boolean,
+
+  /**
+   * Returns whether the events have been loaded or not.
+   */
+  eventsLoaded (): boolean
```
Schedulaar: I think I'd rather call these functions `xyzAvailable`.

src/modules/endpoint/DataContainer.js
```diff
@@ -0,0 +1,250 @@
...
+   */
+  setContext: (cityCode: string, languageCode: string) => Promise<void>,
+
+  /**
+   * Returns whether the current context matches the given city-language combination or not.
+   * @param cityCode
+   * @param languageCode
+   */
+
+  hasContext: (cityCode: string, languageCode: string) => boolean,
```
NKlug: @Schedulaar I added this again for performance reasons. The way we did it without it took about 1 second each time the page was changed.

src/modules/app/components/App.js
```diff
@@ -12,17 +12,17 @@ import type { Store } from 'redux'
 import type { StateType } from '../StateType'
 import type { StoreActionType } from '../StoreActionType'
 import Navigator from './Navigator'
-import MemoryDatabase from '../../endpoint/MemoryDatabase'
+import DataContainer from '../../endpoint/DataContainer'
 
 class App extends React.Component<{}, { waitingForStore: boolean }> {
   store: Store<StateType, StoreActionType>
-  database: MemoryDatabase
+  dataContainer: DataContainer
```
Schedulaar:
```suggestion
  dataContainer: DataContainerInterface
```

src/modules/app/createReduxStore.js
```diff
@@ -23,26 +23,26 @@ import { defaultCitiesState, defaultCityContentState } from './StateType'
...
+import DataContainer from '../endpoint/DataContainer'
 import citiesReducer from '../endpoint/reducers/citiesReducer'
 import watchFetchCategory from '../endpoint/sagas/watchFetchCategory'
 import watchFetchCities from '../endpoint/sagas/watchFetchCities'
 import cityContentReducer from '../endpoint/reducers/cityContentReducer'
 import watchFetchEvent from '../endpoint/sagas/watchFetchEvent'
 import watchContentLanguageSwitch from '../endpoint/sagas/watchContentLanguageSwitch'
 
-function * rootSaga (database: MemoryDatabase): Saga<void> {
+function * rootSaga (dataContainer: DataContainer): Saga<void> {
```
Schedulaar:
```suggestion
function * rootSaga (dataContainer: DataContainerInterface): Saga<void> {
```

src/modules/app/createReduxStore.js
```diff
@@ -23,26 +23,26 @@ import { defaultCitiesState, defaultCityContentState } from './StateType'
...
+    call(watchFetchEvent, dataContainer),
+    call(watchFetchCities, dataContainer),
+    call(watchContentLanguageSwitch, dataContainer),
     call(networkSaga, {})
   ])
 }
 
 const createReduxStore = (
-  database: MemoryDatabase, callback: () => void
+  dataContainer: DataContainer, callback: () => void
```
Schedulaar:
```suggestion
  dataContainer: DataContainerInterface, callback: () => void
```

src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -25,10 +25,9 @@ const createErrorMessage = (fetchResult: FetchResultType) => {
   }, '')
 }
 
-export default function * fetchResourceCache (city: string, language: string, fetchMap: FetchMapType, database: MemoryDatabase): Saga<void> {
-  yield call(database.readResourceCache)
-
-  if (isEmpty(fetchMap)) {
+export default function * fetchResourceCache (
+  city: string, language: string, fetchMap: FetchMapType, dataContainer: DataContainer): Saga<void> {
```
Schedulaar:
```suggestion
  city: string, language: string, fetchMap: FetchMapType, dataContainer: DataContainerInterface): Saga<void> {
```

src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -2,26 +2,25 @@
...
 import loadEvents from './loadEvents'
 import fetchResourceCache from './fetchResourceCache'
 
-export default function * loadCityContent (database: MemoryDatabase, newCity: string, newLanguage: string): Saga<void> {
-  if (database.hasContext(new MemoryDatabaseContext(newCity, newLanguage))) {
-    return // All data is already in the database
+export default function * loadCityContent (dataContainer: DataContainer, newCity: string, newLanguage: string): Saga<void> {
+  if (dataContainer.hasContext(newCity, newLanguage)) {
+    return
   }
```
Schedulaar: I think we should remove this check. It shouldn't harm if we just set the context on the datacontainer even though it's already set.
(Hence we should also remove the function hasContext from dataContainer)

src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -2,26 +2,25 @@
...
+import DataContainer from '../DataContainer'
 import loadLanguages from './loadLanguages'
 import loadCategories from './loadCategories'
 import loadEvents from './loadEvents'
 import fetchResourceCache from './fetchResourceCache'
 
-export default function * loadCityContent (database: MemoryDatabase, newCity: string, newLanguage: string): Saga<void> {
-  if (database.hasContext(new MemoryDatabaseContext(newCity, newLanguage))) {
-    return // All data is already in the database
+export default function * loadCityContent (dataContainer: DataContainer, newCity: string, newLanguage: string): Saga<void> {
```
Schedulaar:
```suggestion
export default function * loadCityContent (dataContainer: DataContainerInterface, newCity: string, newLanguage: string): Saga<void> {
```

src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -25,10 +25,9 @@ const createErrorMessage = (fetchResult: FetchResultType) => {
...
+  if (isEmpty(fetchMap) && dataContainer.resourceCacheLoaded()) {
```
Schedulaar: Also, why should we handle the fetchMap-empty case separately ? May not be part of this task, but I think this whole if block can be removed... Is this only because we haven't implement the native part on iOS @maxammann? Then we could add a comment here maybe...

But why do you need to check here if resourceCache has loaded?

src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -76,8 +75,13 @@ export default function * fetchResourceCache (city: string, language: string, fe
       }, {})
     )
 
-    database.addCacheEntries(resourceCache)
-    yield call(database.writeResourceCache)
+    let newResourceCache = resourceCache
+    if (dataContainer.resourceCacheLoaded()) {
+      const currentResourceCache = yield call(dataContainer.getResourceCache)
+      newResourceCache = {...resourceCache, ...currentResourceCache}
+    }
```
Schedulaar: Yeah, so we might want to add an `addCacheEntries` function in our container...
Overwriting the resourceCache doesn't make sense most of the time, because the files that have previously been downloaded, are still persisted...

src/modules/endpoint/sagas/loadCategories.js
```diff
@@ -8,20 +8,17 @@ import { baseUrl } from '../constants'
...
   const categoriesPayload: CategoriesMapModel = yield call(() => request(createCategoriesEndpoint(baseUrl), params))
   return categoriesPayload.data
 }
 
-function * loadCategories (city: string, language: string, database: MemoryDatabase): Saga<FetchMapType> {
-  // Load data from the disk if existent
-  yield call(database.readCategories)
-
-  if (database.categoriesLoaded()) {
+function * loadCategories (city: string, language: string, dataContainer: DataContainer): Saga<FetchMapType> {
```
Schedulaar:
```suggestion
function * loadCategories (city: string, language: string, dataContainer: DataContainerInterface): Saga<FetchMapType> {
```

src/modules/endpoint/sagas/watchContentLanguageSwitch.js
```diff
@@ -1,24 +1,30 @@
...
+import { all, call, put, takeLatest } from 'redux-saga/effects'
 import type {
   MorphContentLanguageActionType, SwitchContentLanguageActionType, SwitchContentLanguageFailedActionType
 } from '../../app/StoreActionType'
-import MemoryDatabase from '../MemoryDatabase'
+import DataContainer from '../DataContainer'
 import loadCityContent from './loadCityContent'
 
-function * switchContentLanguage (database: MemoryDatabase, action: SwitchContentLanguageActionType): Saga<void> {
+function * switchContentLanguage (dataContainer: DataContainer, action: SwitchContentLanguageActionType): Saga<void> {
```
Schedulaar:
```suggestion
function * switchContentLanguage (dataContainer: DataContainerInterface, action: SwitchContentLanguageActionType): Saga<void> {
```

src/modules/endpoint/sagas/watchFetchCities.js
```diff
@@ -6,23 +6,23 @@ import type {
...
 } from '../../app/StoreActionType'
-import MemoryDatabase from '../MemoryDatabase'
+import DataContainer from '../DataContainer'
 import { createCitiesEndpoint, Payload } from '@integreat-app/integreat-api-client'
 import CityModel from '@integreat-app/integreat-api-client/models/CityModel'
 import request from '../request'
 import { baseUrl } from '../constants'
 
-function * fetchCities (database: MemoryDatabase): Saga<void> {
+function * fetchCities (dataContainer: DataContainer): Saga<void> {
```
Schedulaar:
```suggestion
function * fetchCities (dataContainer: DataContainerInterface): Saga<void> {
```

src/modules/endpoint/sagas/watchFetchCities.js
```diff
@@ -35,6 +35,6 @@ function * fetchCities (database: MemoryDatabase): Saga<void> {
   }
 }
 
-export default function * (database: MemoryDatabase): Saga<void> {
-  yield takeLatest(`FETCH_CITIES`, fetchCities, database)
+export default function * (dataContainer: DataContainer): Saga<void> {
```
Schedulaar:
```suggestion
export default function * (dataContainer: DataContainerInterface): Saga<void> {
```

src/modules/endpoint/sagas/watchFetchEvent.js
```diff
@@ -1,22 +1,28 @@
...
 import type { Saga } from 'redux-saga'
-import { call, put, takeLatest } from 'redux-saga/effects'
+import { all, call, put, takeLatest } from 'redux-saga/effects'
 import type { FetchEventActionType, FetchEventFailedActionType, PushEventActionType } from '../../app/StoreActionType'
-import MemoryDatabase from '../MemoryDatabase'
+import DataContainer from '../DataContainer'
 import loadCityContent from './loadCityContent'
 
-function * fetchEvent (database: MemoryDatabase, action: FetchEventActionType): Saga<void> {
+function * fetchEvent (dataContainer: DataContainer, action: FetchEventActionType): Saga<void> {
```
Schedulaar:
```suggestion
function * fetchEvent (dataContainer: DataContainerInterface, action: FetchEventActionType): Saga<void> {
```

src/modules/endpoint/sagas/watchFetchEvent.js
```diff
@@ -1,22 +1,28 @@
...
 
     const insert: PushEventActionType = {
       type: `PUSH_EVENT`,
       params: {
-        events: database.events,
-        languages: database.languages,
-        resourceCache: database.resourceCache,
+        events: events,
+        languages: languages,
+        resourceCache: resourceCache,
```
Schedulaar: you can use
```js
params: {
  events,
  languages,
  resourceCache,
  ...
}
```
here

src/modules/endpoint/sagas/watchFetchEvent.js
```diff
@@ -34,6 +40,6 @@ function * fetchEvent (database: MemoryDatabase, action: FetchEventActionType):
   }
 }
 
-export default function * (database: MemoryDatabase): Saga<void> {
-  yield takeLatest(`FETCH_EVENT`, fetchEvent, database)
+export default function * (dataContainer: DataContainer): Saga<void> {
```
Schedulaar:
```suggestion
export default function * (dataContainer: DataContainerInterface): Saga<void> {
```

src/modules/endpoint/DataContainer.js
Schedulaar: then there's something wrong with the rest of the code :D

NKlug: The thing is that all the other sagas are executed to the point where the code checks whether the corresponding data has already been loaded or not. I think all the generator-saga-thing just takes a bit.

src/modules/endpoint/sagas/fetchResourceCache.js
NKlug: In general if the fetch map is empty we do not have to run the code below. But if the resourceCache in dataContainer is still `null` we want it to be initialized as `{}`.

src/modules/endpoint/sagas/fetchResourceCache.js
NKlug: I agree, at the moment this would make sense. Do you think `addCacheEntries` should also create an empty ResourceCache if it's `null`?   
On the other hand, as soon as we do incremental updating we probably want to manipulate the ResourceCache here and set the new one manually.

src/modules/endpoint/DataContainer.js
```diff
@@ -0,0 +1,238 @@
...
+import {
+  CategoriesMapModel,
+  CityModel, EventModel,
+  LanguageModel
+} from '@integreat-app/integreat-api-client'
+import DatabaseContext from './DatabaseContext'
+import type { ResourceCacheStateType } from '../app/StateType'
+import DatabaseConnector from './DatabaseConnector'
+
+interface DataContainerInterface {
```
NKlug: @Schedulaar Should we `export default` the interface instead of the implementation. The interface is the thing that will be used almost everywhere in the code and I think one can easily make the mistake and accidentally import the implementation the way it is now.
I personally like `DataContainer` better as a name for the interface and call the implementation something like `DefaultDataContainer` or something similar.

src/modules/endpoint/DataContainer.js
NKlug: Nvm, the wrong `xyzLoaded()` functions caused the delay as discussed.

src/modules/endpoint/DataContainer.js
NKlug: Yep, sound better. Done.

src/modules/endpoint/DataContainer.js
Schedulaar: I have no preference here :thinking:

src/modules/endpoint/sagas/fetchResourceCache.js
Schedulaar: > I agree, at the moment this would make sense. Do you think addCacheEntries should also create an empty ResourceCache if it's null?

Yes would make sense.

> On the other hand, as soon as we do incremental updating we probably want to manipulate the ResourceCache here and set the new one manually

I'd say thats too far away...

src/modules/endpoint/DataContainer.js
```diff
@@ -0,0 +1,238 @@
...
+  _resourceCache: ResourceCacheStateType | null
+  _events: Array<EventModel> | null
+
+  constructor () {
+    this._databaseConnector = new DatabaseConnector()
+  }
+
+  getCities = async (): Promise<Array<CityModel>> => {
+    if (this._cities === null) {
+      throw Error('Cities are null.')
```
maxammann: I think we can continue to live with the exceptions but this is not a perfect design. By perfect I mean that it is not possible to miss-use the class.

maxammann:
A fix for builds is available here: https://github.com/Integreat/integreat-react-native-app/pull/41#pullrequestreview-228365870

NKlug:
`DataContainer` is more like a cache for the data obtained by calling methods of `DatabaseConnector`. I agree that `DataContainer` isn't the best name, although I'm not so content about `LazyDatabase` either. Maybe just `DataCache`?
I disagree with calling the `DatabaseConnector` database, since it just provides access to the real database (if you'd like to call the json files that ^^) and holds no data itself.
Other opinions @Schedulaar ?

src/modules/endpoint/DataContainer.js
Schedulaar: I was unable to think of a better design though :/

Schedulaar:
Sure, there could be "confusions" with `redux`-containers, but still that's what it is: https://en.wikipedia.org/wiki/Container_(abstract_data_type)

I'd vote for `DataContainer`. I'd rather replace the first word (`Data`) but I can't think of a better one^^

src/modules/endpoint/DataContainer.js
NKlug: I implemented it as suggested.

src/modules/endpoint/sagas/fetchResourceCache.js
NKlug: Also if there is nothing to fetch there is no need to run any native code. In theory this affects the performance although I don't know to which extent.

src/modules/endpoint/sagas/fetchResourceCache.js
NKlug: Done.

src/modules/endpoint/DataContainer.js
```diff
@@ -0,0 +1,238 @@
...
+import {
+  CategoriesMapModel,
+  CityModel, EventModel,
+  LanguageModel
+} from '@integreat-app/integreat-api-client'
+import DatabaseContext from './DatabaseContext'
+import type { ResourceCacheStateType } from '../app/StateType'
+import DatabaseConnector from './DatabaseConnector'
+
+export interface DataContainerInterface {
```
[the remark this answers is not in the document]
NKlug: Renamed and moved to a separate file.

PR #42, opened 2019-04-18
NATIVE-108: Disable language buttons if not available

klinzo:


src/modules/layout/containers/HeaderContainer.js
```diff
@@ -2,6 +2,23 @@
...
+import type { StateType } from '../../app/StateType'
+import { connect } from 'react-redux'
+import React from 'react'
+import { withNavigation } from 'react-navigation'
 
-// $FlowFixMe
-export default withTheme(Header)
+const mapStateToProps = (state: StateType, ownProps) => {
+  const key: string = ownProps.navigation.getParam('key')
+  // todo add availableLanguages for other routes
```
klinzo: Not quite sure if I should do this in this task, since there is https://issues.integreat-app.de/browse/NATIVE-180 .

src/modules/common/components/Selector.js
```diff
@@ -49,19 +43,34 @@ export const Wrapper = styled.View`
...
 class Selector extends React.Component<PropsType> {
   getItems (): React.Node {
-    const {items, activeItemCode} = this.props
+    const {items, selectedItemCode} = this.props
     return items.map(item => {
+      const isSelected = item.code === selectedItemCode
+      if (item.active || isSelected) {
+        return (
+          <TouchTarget key={item.code} onPress={item.onPress}>
+            <ActiveElement selected={isSelected}>
```
NKlug: Is the `theme` passed automatically?

src/modules/layout/components/Header.js
```diff
@@ -144,7 +152,8 @@ class Header extends React.PureComponent<PropsType, StateType> {
           </HorizontalLeft>
           <MaterialHeaderButtons>
             <Item title='Search' iconName='search' onPress={this.showSearchBar} />
-            <Item title='Change Language' iconName='language' onPress={this.goToLanguageChange} />
+            <Item title='Change Language' iconName='language'
+                  onPress={isOtherLanguageAvailable ? this.goToLanguageChange : null} />
```
NKlug: Should we display any kind of information that there is no translation for the current page?

src/modules/common/components/Selector.js
maxammann: Theme should be passed explicitly so this is easier to test.

src/modules/layout/components/Header.js
maxammann: I think it is fine if the button is greyed-out and not clickable. At least I dont know how to give proper feedback.
Maybe a shaking-animation: http://www.mircozeiss.com/shake-that-login-form-with-angularjs

but this it not important and can be done in a separate task

src/modules/layout/containers/HeaderContainer.js
```diff
@@ -2,6 +2,23 @@
...
+  const targetRoute = state.cityContent.categoriesRouteMapping[key]
+  const model = targetRoute && targetRoute.root && targetRoute.models[targetRoute.root]
+  const cityLanguages = state.cityContent.languages && state.cityContent.languages.map(languageModel => languageModel.code)
+  const availableLanguages = model && model.title !== state.cityContent.city ? Array.from(model.availableLanguages.keys()) : cityLanguages
```
maxammann: I think we should create a helper for stuff getting stuff like this. We have the reselect library which allows you to write selectors: https://github.com/reduxjs/reselect

This is not needed so far but I think it would make sense to write a function to query the store for values by route key.

src/modules/layout/containers/HeaderContainer.js
maxammann: This should work for events and categories for now.

src/modules/layout/containers/HeaderContainer.js
```diff
@@ -2,6 +2,23 @@
...
 import { withTheme } from 'styled-components/native'
+import type { StateType } from '../../app/StateType'
+import { connect } from 'react-redux'
+import React from 'react'
+import { withNavigation } from 'react-navigation'
 
-// $FlowFixMe
-export default withTheme(Header)
+const mapStateToProps = (state: StateType, ownProps) => {
+  const key: string = ownProps.navigation.getParam('key')
```
maxammann: This is the `key` from the navigation actions which lead to the dashboard/categories or events, right?

maxammann:
A fix for builds is available here: https://github.com/Integreat/integreat-react-native-app/pull/41#pullrequestreview-228365870

src/modules/layout/components/Header.js
```diff
@@ -120,22 +121,29 @@ class Header extends React.PureComponent<PropsType, StateType> {
...
+      return <BoxShadow theme={theme}><HorizontalLeft>
         <HeaderBackButton onPress={this.closeSearchBar} />
-        <ThemedSearchBar theme={this.props.theme} />
+        <ThemedSearchBar theme={theme} />
       </HorizontalLeft>
       </BoxShadow>
     }
 
     const headerTitle = this.getDescriptor().headerTitle || ''
+    const isOtherLanguageAvailable = availableLanguages && availableLanguages.length > 0
```
Schedulaar: why do you need to check for `availableLanguages` to be truthy? The proptype suggests it to be always truthy.

src/modules/layout/components/Header.js
Schedulaar: I think we could also just not care if no other language is available and still open the modal -- I guess this is a rare case...

src/modules/layout/containers/HeaderContainer.js
Schedulaar: Still, I think using `model.title` to check if `model` represents the root (at least that's what I think you want to do) is not good. Can't you use `model.isRoot()` instead?

src/modules/layout/components/Header.js
klinzo: I don't care. It could also be greyed out and show a toast on click.

src/modules/layout/components/Header.js
klinzo: The prop types are wrong, it isn't always truthy.

src/modules/common/components/Selector.js
klinzo: To which components does the theme has to be passed explicitly?

src/modules/layout/components/Header.js
klinzo: Alright I just removed it and we won't care.

src/modules/common/components/Selector.js
maxammann: Basically every component should have the `theme` in the props. Only containers should be able to get the theme by using `withTheme`

src/modules/layout/containers/HeaderContainer.js
klinzo: @maxammann using the reselect library now to select the availableLanguages. Could you please take a look if this is allright and if the location in `modules/common/selectors/` is okay?

src/modules/layout/containers/HeaderContainer.js
klinzo: Probably, yes. Should be the key of the current screen

src/modules/common/components/Selector.js
klinzo: Even if they are not "real" react components but just styled components?

maxammann: Imo yes as it makes testing easier and is not much work. At least I would just change the code and pass the theme prop instead of making the test more complicated.

src/modules/layout/containers/HeaderContainer.js
maxammann: Looks fine :+1:

src/modules/common/models/SelectorItemModel.js
```diff
@@ -2,11 +2,13 @@
 export default class SelectorItemModel {
   _code: string
   _name: string
+  _active: boolean
```
Schedulaar: I'd rather go for the term `enabled` here which is much more common (although I think we had this discussion in the webapp already^^)
https://developer.apple.com/documentation/uikit/uicontrol/1618217-enabled?language=objc
https://docs.microsoft.com/en-us/dotnet/api/system.web.ui.webcontrols.webcontrol.enabled?view=netframework-4.8#remarks

src/modules/common/selectors/availableLanguagesSelector.js
```diff
@@ -0,0 +1,42 @@
...
+  categoryRouteSelector,
+  (route: ?CategoryRouteStateType): ?CategoryModel => route && route.root && route.models[route.root]
+)
+
+const eventModelSelector = createSelector(
+  eventRouteSelector,
+  (route: ?EventRouteStateType): ?EventModel => route && route.path && route.models[route.path]
+)
+
+export const availableLanguagesSelector = createSelector(
```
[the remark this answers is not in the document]
klinzo: Jep, nice improvement. Will wait for that PR to be merged.

src/modules/common/models/SelectorItemModel.js
klinzo: Okay for me.

maxammann:
@klinzo should I help you with the failing build? Else I think this is rdy for merge :)

PR #44, opened 2019-04-23
NATIVE-83: Detect language and E2E tests

maxammann:


maxammann:
ping

e2e/tests/modules/i18n/ui-language.e2e.js
```diff
@@ -0,0 +1,32 @@
...
+  it('should match the system language de', async () => {
+    const driver = await setupDriver(select({
+      android: {language: 'de', locale: 'DE'},
+      ios: {language: 'de', locale: 'de_DE'}
+    }))
+
+    const landingPage = new LandingPage(driver)
+    await landingPage.ready()
+
+    expect(await landingPage.getSearchInput().text()).toBe('Suche nach deinem Ort')
```
NKlug: I don't really like comparing strings here. IMO the a test should not fail only because I changed something in the `locales.json`. Same below.

maxammann: I agree but there is not really a other way to do this. E2E tests can fail if you change the visuals of the app. I don't think we should make 100% sure they are as stable as unit tests. Also E2E tests should not depend on the source code imo. So there is not really a way to change that.

maxammann:
ping

android/app/src/main/java/com/integreat/MainActivity.java
```diff
@@ -12,4 +24,15 @@
...
     }
+
+    @Override
+    public void onConfigurationChanged(Configuration newConfig) {
+        super.onConfigurationChanged(newConfig);
+
+        Locale locale = newConfig.locale;
+        if (!currentLocale.getISO3Language().equals(locale.getISO3Language())) {
+            currentLocale = locale;
+            getReactInstanceManager().recreateReactContextInBackground();
```
Schedulaar: So what does that do again? "Reload" the entire app and forget about the previous state?

src/modules/i18n/containers/I18nProvider.js
```diff
@@ -84,25 +84,16 @@ export class I18nProvider extends React.Component<PropsType, {
...
   componentDidMount () {
-    this.setLanguage(this.props.language)
-  }
-
-  componentDidUpdate (prevProps: PropsType) {
-    if (this.props.language !== prevProps.language) {
-      this.setLanguage(this.props.language)
-    }
+    this.initLanguage(this.props.language)
   }
```
Schedulaar: Why did you remove the `componentDidUpdate` thing?
We would not set the ui direction in the on language change this way and therefore the ui-direction does not correspond to `cityContent.language` in the store (but to the systemLanguage?).
I think we should have a stricter separation of content-language (which is the one in the store at cityContent.language I guess) and system language...
If you want to only take care about the system-language in the I18nProvider, then please remove the mapStateToProps here and rename all `language` occurences to `systemLanguage`.

Schedulaar: Also we need to think about what we should do with the fonts, but thats another task...

android/app/src/main/java/com/integreat/MainActivity.java
maxammann: Yes, else the app stays in the old language on Android. Now it behaves the same on Android and iOS

src/modules/i18n/containers/I18nProvider.js
maxammann: I think we should do this in a separate task. The reason why I renamed setLanguage to initLanguage is that during the runtime of the app the system language will never change.

Also setUiDirection is not doing anything right now and we should probably remove it. But we have to dig deeper into RN to make sure it works as expected: https://facebook.github.io/react-native/blog/2016/08/19/right-to-left-support-for-react-native-apps

maxammann: https://issues.integreat-app.de/browse/NATIVE-191

PR #46, opened 2019-05-03
NATIVE-109: Enable linking to pages in the same city and language

klinzo:


src/modules/common/components/Page.js
```diff
@@ -38,13 +38,16 @@ type PropType = {
...
   navigation: NavigationScreenProp<*>,
+  navigateToIntegreatUrl: (url: string, cityCode: string, language: string) => void,
   files: FileCacheStateType,
   children?: React.Node,
   language: string,
   cityCode: string,
   lastUpdate: Moment
 }
 
+const HIJACK = /https?:\/\/(cms(-test)?\.integreat-app\.de|web\.integreat-app\.de|integreat\.app)(?!\/[^/]*\/(wp-content|wp-admin|wp-json)\/.*).*/
```
klinzo: Not quite sure if I should leave the cms-test capture.

index.js
```diff
@@ -1,12 +1,18 @@
...
 import 'moment/locale/de' // fixme
 
 // @see: https://github.com/facebook/react-native/issues/9599
 if (typeof global.self === 'undefined') {
   global.self = global
 }
 
+// @see https://github.com/facebook/react-native/issues/16434
+global.Buffer = Buffer
+global.URL = URL
```
maxammann: Are you sure you need to override the current one in the global scope?
As far as I understand the usual URL implementation is missing. So can you just import URL in `createNavigateToIntegreatUrl.js` and use the imported one?

I think we do not need to override this one.

index.js
```diff
@@ -1,12 +1,18 @@
...
+import { URL } from 'whatwg-url'
 import 'moment/locale/de' // fixme
 
 // @see: https://github.com/facebook/react-native/issues/9599
 if (typeof global.self === 'undefined') {
   global.self = global
 }
 
+// @see https://github.com/facebook/react-native/issues/16434
+global.Buffer = Buffer
```
maxammann: Do we need the Buffer?

src/modules/common/components/Page.js
maxammann: I think it should stay because without it content created in the cms-test would not have internal linking.

src/modules/app/createNavigateToIntegreatUrl.js
```diff
@@ -0,0 +1,31 @@
...
+import type { Dispatch } from 'redux'
+import type { NavigationScreenProp } from 'react-navigation'
+import createNavigateToEvent from './createNavigateToEvent'
+
+export default (dispatch: Dispatch<*>, navigation: NavigationScreenProp<*>) =>
+  (url: string, cityCode: string, language: string) => {
+    const pathname = decodeURIComponent(new URL(decodeURIComponent(url)).pathname)
+    const parts = pathname.split('/').filter(segment => segment)
+
+    if (parts[0] === cityCode && parts[1] === language) {
```
Schedulaar: `/augsburg` should open the dashboard in the current language as well

src/modules/app/createNavigateToIntegreatUrl.js
```diff
@@ -0,0 +1,31 @@
...
+          createNavigateToEvent(dispatch, navigation)(cityCode, language)
+        }
+      } else if (parts[2]) {
+        // '/augsburg/de/willkommen'
+        createNavigateToCategory('Categories', dispatch, navigation)(cityCode, language, pathname)
+      } else {
+        // '/augsburg/de'
+        createNavigateToCategory('Dashboard', dispatch, navigation)(cityCode, language, pathname)
+      }
+    }
```
Schedulaar: Can you add a todo for the other cases?

index.js
klinzo: Maybe just importing URL works, but afaik buffer has to be set globally, because URL uses it internally.

index.js
klinzo: Yep.

src/modules/app/createNavigateToIntegreatUrl.js
klinzo: Aren't there tasks for this?

PR #47, opened 2019-05-03
NATIVE-107: Language change with untranslatable stack

Schedulaar:


src/locales.json
```diff
@@ -77,6 +77,16 @@
       "lastUpdate": "آخرین به روزآوری: "
     }
   },
+  "common": {
+    "en": {
+      "notAvailable": "Not Available",
```
maxammann: This key is already present in the locales.json as `NotFoundLanguage`: Sorry, but your language is not available here. Also see `ChooseALanguage`

https://docs.google.com/spreadsheets/d/1kMxVlqzwjjDoARO5Ss5oMNUEjqubT0Ps8NB7cguP-yU/edit#gid=194836756

src/modules/endpoint/reducers/__tests__/switchLanguage.spec.js
```diff
@@ -270,14 +270,7 @@ describe('switchLangauge', () => {
   ]
 
   const initialState: CityContentStateType = {
```
maxammann: If you want you can rename this test to morthLanguage.spec.js

src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -49,5 +52,12 @@ const mapStateToProps = (state: StateType, ownProps) => {
 
 // $FlowFixMe
 const themed = withTheme(props => <ScrollView><Categories {...props} /></ScrollView>)
-// $FlowFixMe connect()
-export default withRouteCleaner(connect(mapStateToProps, mapDispatchToProps)(themed))
+export default compose( // $FlowFixMe connect()
```
maxammann: Is that $FlowFixMe still supressing the same issue as before? I'm a bit concerned because of the branch thing because Flow fails a lot with it. If that is the case just add an other $FlowFixMe with a link to a issue maybe.

src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -49,5 +52,12 @@ const mapStateToProps = (state: StateType, ownProps) => {
...
+  withRouteCleaner,
+  connect((state: StateType, ownProps) => ({
+    invalidLanguage: !state.cityContent.categoriesRouteMapping[ownProps.navigation.getParam('key')]
```
maxammann: I think it would make sense to create a function for `ownProps.navigation.getParam('key')` at some point.

src/routes/events/containers/EventsContainer.js
```diff
@@ -39,6 +41,11 @@ const mapDispatchToProps = (dispatch: Dispatch<StoreActionType>, ownProps) => ({
 
 export default compose(
   translate('events'),
+  connect((state: StateType, ownProps) => ({
```
maxammann: Same as above: Concerned about types :)

src/locales.json
Schedulaar: Right, I should have checked that. I replaced the keys.

src/modules/endpoint/reducers/__tests__/switchLanguage.spec.js
Schedulaar: Done.

src/routes/categories/containers/CategoriesContainer.js
Schedulaar: I could remove all flowfixmes.

src/routes/events/containers/EventsContainer.js
Schedulaar: What exactly do you want me to do here? :thinking:

src/modules/common/components/PageNotAvailable.js
```diff
@@ -0,0 +1,51 @@
...
+
+class PageNotAvailable extends React.Component<PropsType> {
+  onPress = (model: LanguageModel) => {
+    this.props.changeLanguage(this.props.city, model.code)
+  }
+
+  render () {
+    const {t, languages, theme} = this.props
+    return <Wrapper theme={theme}>
+      <Caption title={t('notFoundLanguage')} />
```
klinzo: Kind of conflicting with the component name `PageNotAvailable`. Either this should be `LanguageNotAvailablePage` (or something similar) or the caption should not be constant.

src/modules/endpoint/reducers/morphContentLanguage.js
```diff
@@ -1,83 +1,83 @@
...
+      const translatedRoot = allAvailableLanguages.get(newLanguage)
 
-    const translatedArray = reduce(children[path], (result, key) => {
-      const translatedKey = models[key].availableLanguages.get(newLanguage)
-      if (!translatedKey) {
-        // TODO: This is code for debugging which could help in the future. Remove once this has been tested in NATIVE-116
-        console.warn(`Path ${key} is not translatable!`)
-        return result
+      if (!translatedRoot) { // Route is not translatable
+        return {...route}
```
klinzo: Why not just returning `route`?

src/modules/endpoint/reducers/morphContentLanguage.js
```diff
@@ -1,83 +1,83 @@
...
 
-      if (!newCategoriesMap.findCategoryByPath(translatedKey)) {
-        // TODO: This is code for debugging which could help in the future. Remove once this has been tested in NATIVE-116
-        console.warn(`Path ${translatedKey} does not exist in new model!`)
-        return result
+      const rootModel = newCategoriesMap.findCategoryByPath(translatedRoot)
+      if (!rootModel) {
+        console.warn(`Inconsistent data detected: ${translatedRoot} does not exist,
+                      but is referenced in ${root} as translation for ${newLanguage}.`)
+        return {...route}
```
klinzo: Same again.

src/modules/endpoint/reducers/morphContentLanguage.js
```diff
@@ -1,83 +1,83 @@
...
+import CategoriesMapModel from '@integreat-app/integreat-api-client/models/CategoriesMapModel'
+import forEachTreeNode from '../../common/forEachTreeNode'
+import { isEmpty } from 'lodash'
 
-    if (!translatedKey) {
-      // TODO: This is code for debugging which could help in the future. Remove once this has been tested in NATIVE-116
-      console.warn(`Path ${path} is not translatable!`)
-      return result
-    }
+const categoryRouteTranslator = (newCategoriesMap: CategoriesMapModel, city: string, newLanguage: string) =>
```
klinzo: I think city is not required here, is it?

src/modules/endpoint/reducers/morphContentLanguage.js
```diff
@@ -1,83 +1,83 @@
...
+    if (!path) { // Route is a list of all events
+      return {path: null, models: newEvents, allAvailableLanguages}
+    }
+    // Route is a single event
 
-const translateModels = (models, newCategoriesMap, currentCity, newLanguage) =>
-  reduce(models, (result, value: CategoryModel) => {
-    const translatedKey = translatePath(value, currentCity, newLanguage)
+    if (isEmpty(models)) {
+      throw new Error(`No model for route ${navKey}!`)
```
klinzo: Is this check required?

src/modules/endpoint/reducers/morphContentLanguage.js
klinzo: Also, does it make sense to return the untranslated route? Shouldn't we rather throw or return null?

src/routes/events/containers/EventsContainer.js
maxammann: As you resolved all flowfixme this is fine.

src/modules/endpoint/reducers/morphContentLanguage.js
Schedulaar: > Why not just returning route?

True :D

>

Also, does it make sense to return the untranslated route? Shouldn't we rather throw or return null?

The important thing is that the `allAvailableLanguages` is still given for the route (to see, whether the route contains valid information for this language and, if not, which languages are available for the route.

src/modules/endpoint/reducers/morphContentLanguage.js
Schedulaar: No, removed.

src/modules/endpoint/reducers/morphContentLanguage.js
Schedulaar: No, thanks. Removed.

src/modules/endpoint/reducers/morphContentLanguage.js
Schedulaar: Done.

src/modules/common/components/PageNotAvailable.js
Schedulaar: Renamed the locale key to languageNotAvailable and the component to LanguageNotAvailablePage.

src/modules/endpoint/reducers/morphContentLanguage.js
klinzo: There were some more `{...route}` did you also fix  those?

PR #53, opened 2019-05-08
NATIVE-205: Make caption use textColor of theme

Schedulaar:


src/modules/sprungbrett/components/SprungbrettExtra.js
```diff
@@ -36,7 +38,7 @@ class SprungbrettExtra extends React.Component<PropsType> {
 
     return (
       <>
-        <Caption title={extra.title} />
+        <Caption title={extra.title} theme={this.props.theme} />
```
NKlug: For consistency extract `theme` in line 32.

PR #54, opened 2019-05-08
NATIVE-204: New navigation tiles

Schedulaar:


src/modules/common/components/NavigationTiles.js
```diff
@@ -13,16 +13,12 @@ type PropsType = {|
...
   display: flex;
   flex-direction: row;
   align-items: center;
   justify-content: space-around;
-`
-const Separator = styled.View`
-  border-bottom-width: 1px;
-  border-bottom-color: ${props => props.theme.colors.themeColor};
-  margin: 0 15px;
+  elevation: 1;
```
NKlug: Elevation does not work on IOS, we have to set the shadow in a different way there.

Schedulaar: I added the `shadow-*` properties for iOS with values that I hope are fine. (I couldn't test this).
image

Jenkinsfile
```diff
@@ -36,12 +36,12 @@ pipeline {
...
                                 E2E_TEST_IDS = "1"
                                 RCT_NO_LAUNCH_PACKAGER = "true"
-                                BUNDLE_CONFIG = "./metro.config.release.js"
+                                BUNDLE_CONFIG = "./metro.config.ci.js"
                             }
                             steps {
                                 lock('pod-compilation') {
-                                    // We are locking 2 steps here beacuse:
-                                    // 1)   "pod install" can not run paralell because pod does not support this
+                                    // We are locking 2 steps here because:
```
maxammann: Haha lolz, yeah I can remember that I was in a hurry when I wrote this :D

metro.config.ci.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  transformer: {
+    getTransformOptions: async () => ({
+      transform: {
+        experimentalImportSupport: false,
+        inlineRequires: false
```
maxammann: Are you sure this is already used? https://facebook.github.io/react-native/blog/2019/03/12/releasing-react-native-059#faster-app-launches-with-inline-requires

Not sure which side-effects this has

Schedulaar: Ahh, no then it's not used yet. I'll remove it...

PR #56, opened 2019-05-09
NATIVE-187: Fix language not available

klinzo:


src/modules/common/containers/LanguageNotAvailableContainer.js
```diff
@@ -0,0 +1,36 @@
...
+
+import { connect } from 'react-redux'
+import type { Dispatch } from 'redux'
+import type { StateType } from '../../../modules/app/StateType'
+import type { StoreActionType } from '../../../modules/app/StoreActionType'
+import LanguageNotAvailablePage from '../../../modules/common/components/LanguageNotAvailablePage'
+import { withTheme } from 'styled-components/native'
+import compose from 'lodash/fp/compose'
+import { translate } from 'react-i18next'
+import createNavigateToCategory from '../../app/createNavigateToCategory'
```
a-kud: I suggest to keep imports in the following order:

1. packages
2. components
3. actions
4. constants
5. flow types
It increases readability and allows you to faster grasp what's imported.
```
import compose from 'lodash/fp/compose'
import { connect } from 'react-redux'
import { translate } from 'react-i18next'

import LanguageNotAvailablePage from '../../../modules/common/components/LanguageNotAvailablePage'
import createNavigateToCategory from '../../app/createNavigateToCategory'

import type { Dispatch } from 'redux'
import type { StateType } from '../../../modules/app/StateType'
import type { StoreActionType } from '../../../modules/app/StoreActionType'
```

src/modules/endpoint/sagas/watchFetchLanguagesForCategory.js
```diff
@@ -0,0 +1,51 @@
...
+      params: {
+        languages,
+        city,
+        language
+      }
+    }
+    yield put(insert)
+
+    if (languages.map(language => language.code).includes(language)) {
+      const fetchCategory: FetchCategoryActionType = {
```
a-kud: better put this action in separate file like ` src/modules/endpoint/actions/index.js` and import them here

src/routes/dashboard/containers/DashboardContainer.js
```diff
@@ -66,7 +69,16 @@ const mapStateToProps = (state: StateType, ownProps) => {
...
-// $FlowFixMe
 const themed = withTheme(Dashboard)
-// $FlowFixMe connect()
-export default withRouteCleaner(connect(mapStateToProps, mapDispatchToProps)(withError(translate('dashboard')(themed))))
+export default compose([
+  withRouteCleaner,
+  connect((state: StateType): { invalidLanguage: boolean } => {
+    const languages = state.cityContent.languages
+    const language = state.cityContent.language
+    return { invalidLanguage: !!languages && !languages.find(languageModel => languageModel.code === language) }
```
a-kud: what's the purpose here?

src/modules/endpoint/sagas/watchFetchLanguagesForCategory.js
Schedulaar: I guess we don't do that for the whole project @a-kud… It's also not too bad, since we use our FlowTypes to create actions, so they're at least well-shaped.
We could discuss if we should add action-creators (which I would rather decline), but that's sth for a separate task...

src/routes/dashboard/containers/DashboardContainer.js
Schedulaar: He checks in the state if the currentLanguage is available and sets the `invalidLanguage` prop accordingly. In the `branch` (next line), he checks whether he should display a LanguageNotAvailableContainer or not.

src/modules/common/containers/LanguageNotAvailableContainer.js
klinzo: Would be a separate task imo since we don't do this anywhere yet, feel free to create a new task.

PR #57, opened 2019-05-09
NATIVE-111: Linking to other languages

klinzo:


src/modules/app/StoreActionType.js
```diff
@@ -97,11 +101,18 @@ export type MorphContentLanguageActionType = {|
   |}
 |}
 
+export type PushContentLanguageActionType = {|
+  type: 'PUSH_CONTENT_LANGUAGE', params: {|
```
maxammann: Use `SET_CONTENT_LANGUAGE` here. I chose `PUSH` to have a counter-action for `FETCH`.

So only name it `PUSH` if there is a corresponding `FETCH`

src/modules/app/createNavigateToCategory.js
```diff
@@ -19,13 +21,11 @@ export default (
       key,
       onRouteClose: () => dispatch({type: 'CLEAR_CATEGORY', params: {key}})
     },
-    // Use key when navigating to Categories. Else a new route is not pushed
-    key: routeName === 'Categories' ? key : undefined /* Explanation for "key":
-                                                         https://reactnavigation.org/docs/en/navigation-key.html */
+    key
```
maxammann: Are you sure the navigator is still working with this? I had the problem that it was sometimes not pushing a new screen when it should I think.

src/modules/app/StateType.js
```diff
@@ -14,13 +14,15 @@ export type CategoryRouteStateType = {|
                                                    (e.g. CategoriesStateType) to save memory
                                                    in the state. This would be an optimization! */
   +children: { [path: PathType]: Array<PathType> },
-  +allAvailableLanguages: Map<string, string> // including the current content language
+  +allAvailableLanguages: Map<string, string>, // including the current content language
+  +previousLanguage: string
```
maxammann: I did not really get why `previousLanguage` is needed. Can you explain this?

Schedulaar: I guess it's used to change the language back to the previous one, when navigating back (see clearCategory)

src/modules/app/createNavigateToCategory.js
klinzo: I didn't notice any problems, could you describe where/when problems occured? Or maybe try it yourself?

klinzo: Apart from that, I don't see a reason why navigation to the Dashboard route wouldn't work if you supply a key.

klinzo: What is not working though is navigating back to the landing screen.

klinzo:
@maxammann @Schedulaar @NKlug Does anyone of you have a clue why the `componentWillUnmount` method of `withRouteCleaner` is suddenly called twice just because I supply a key in `createNavigateToCategory` on navigation to the Dashboard? It's kind of really strange. The key of the route which the route cleaner tries to close is of the pattern `{Date.now}-{uuid}` instead of our `category-id-{Date.now}-{uuid}`, so it looks like the react navigator firstly uses an auto generated key. Also, onRouteClose is for some reason undefined because of which the RouteCleaner throws an error.

klinzo:
Without the key for the Dashboard route this doesn't happen, but if navigating to a different language the old routes are cleared from the state.

src/modules/app/createNavigateToCategory.js
maxammann: That is expected behavior. The landing screen should only be displayed once in the future.

maxammann: I think I did it like this because of the behavior you mention in this PR: https://github.com/Integreat/integreat-react-native-app/pull/57#issuecomment-491813723

Apart from that I don't have clue.

klinzo: So there will be no possibility to change your location afterwards? Apart from that, is my solution by just not throwing errors in routeCleaner? Everything works like it should now.

Schedulaar: Have a look at my PR for pullToRefresh (https://github.com/Integreat/integreat-react-native-app/pull/59) I've fixed it there.

src/modules/endpoint/hocs/withRouteCleaner.js
```diff
@@ -15,10 +15,9 @@ function withRouteCleaner<Props: { navigation: NavigationScreenProp<*> }> (
   class RouteCleaner extends React.PureComponent<Props> {
     componentWillUnmount () {
       const onRouteClose = this.props.navigation.getParam('onRouteClose')
-      if (!onRouteClose) {
-        throw new Error('onRouteClose is not provided to route in the navigation props!')
```
Schedulaar: why did you remove the Error?

src/modules/app/createNavigateToIntegreatUrl.js
```diff
@@ -4,32 +4,51 @@ import createNavigateToCategory from './createNavigateToCategory'
...
+      // same city
+
+      if (newLanguage !== language) {
+        const setContentLanguage: SetContentLanguageActionType = {
+          type: 'SET_CONTENT_LANGUAGE',
+          params: {
+            newLanguage
+          }
+        }
+        dispatch(setContentLanguage)
```
Schedulaar: I'm concerned of the `state.cityContent.language`now: I think it's not at all clear, what this value holds now: Does it hold the "current language of the screen"?
Also what happens with it when navigating back (and potentially changing the "current language of the screen")? I think it would be a lot easier to just switch the content-language (as the SWITCH_CONTENT_LANGUAGE action does) and then push the new page on top... (so the whole stack has the same language)

Edit: I saw now, that you handled this "moving-back"-case. Still, this somehow bugs me...

klinzo: I don't really like this approach. If a category/event in a different language is linked and the user would navigate back, the previous category would also be in the new language, which isn't expected by users imo.

src/modules/app/createNavigateToCategory.js
klinzo: Looks good, I'll wait until your PR is merged.

src/modules/endpoint/hocs/withRouteCleaner.js
klinzo: Reverted

src/modules/app/createNavigateToIntegreatUrl.js
Schedulaar: Yes, that's true; I think as a user, it makes more sense to have the original language when navigating back... On the other hand this increases complexity a lot (having multiple languages opened at the same time) for a use-case that should not occur (because the links deployed by the cms should target the same language)... :thinking:

klinzo: So why do we even need this? This task is solely about linking to stuff in a different language. If there are only links in the same language, this is not needed at all.

maxammann: We decided to do this because we did not really know what we should do in case of a link like: `/ausgburg/de/gesundheit` -> `/ausgburg/en/first-steps`.

I'm very sure that that navigating back should always yield to the page you previously saw. So if a city decides to link to a page in an other language, e.g. de -> ar, then navigating back should bring you back to the de page.

There is actually a usecase for this if a page only exists in one particular language.

TL;DR: We should not ignore the language in a link. Navigating back should yield the previous page and therefore change the language.

I'm not deep into the code, but that is the way is is solved right now, right?

Schedulaar: Well, the problem is, there are links targeting different languages. Although this is a problem of our cms, we need to somehow react adequately to that.

Schedulaar: @maxammann Alright, but then we should consider removing the language as a property in the `state.cityContent.language`, since there's no real concept behind that, and instead just have a language per screen...

maxammann: @Schedulaar Most links are probably correct in the CMS. Handling it the way I described is at least the "right" way to do it. So yes you are right language is an attribute of the current page.

BUT: I could live if it is like @Schedulaar described in the first post for now.


@klinzo It's up to you how you want to solve it:

1. Make language an attribute of the current route
2. Just switch the content-language (as the SWITCH_CONTENT_LANGUAGE action does) and then push the new page on top... (so the whole stack has the same language)

I think the concept of a previousLanguage is not very clean and you probably have an issue in the following case: You follow the links: `/augsburg/de/` -> `/augsburg/en/` -> `/augsburg/ar/`.

If you navigate back now 2 times the "previousLanguage" de is lost.

klinzo: @maxammann No, the previous language would not be lost if navigating back 2 times.

klinzo: I think we should discuss this in person.

maxammann: Alright, Tuesday?

Schedulaar: So I thought about this once again:
We also need to consider what happens when the user follows a link to a different language and then explicitly switches the language on that page (Also: Which language should the modal display as `selected`? I think the current language of the page).
Imo the following concept would make sense if we want to support this navigating-back-preserves-language feature:

We maintain the `state.cityContent.language´ not as the current-screen-language, but the "global" setting of the current contentLanguage (so it does not change when following a link to a different language). When changing a language explicitly by the user, this global language will be switched.
Other than that we should maintain the current language per route/screen. As far as I can see, we would not need to remember this `previousLanguage` that way.
Using this, we need to somehow give the `LanguageModal` the current language of the screen, which should technically be possible afaik.

maxammann: I think @Schedulaar s post makes a lot of sense. Ofc this is a complicated solution but behaves like I would expect.

klinzo: So if the language is switched explicitly by the user, all route language props are switched to the newLanugae?

klinzo:
@Schedulaar @maxammann The only thing still missing is the displaying of the right selected language in the `ChangeLanguageModal`. The problem here is that it isn't really easy to get the right route key of the underlying route (since the modal has its own). If there isn't an easy solution for this that you know of, I would just create a task for this.

Schedulaar:
@klinzo Couldn't you put it in the params of the navigation action?
I assume we navigate to the ChangeLanguageModal in the Header. There we have access to the key of the current screen...

klinzo:
@Schedulaar Ähheeeeeeem....noooooo, definitely not possible?!? :grin: Would be way to easy. To be serious: Yeah, should work I'm stupid ^^ :man_facepalming:

src/routes/events/containers/EventsContainer.js
```diff
@@ -45,7 +45,9 @@ export default compose([
   translate('events'),
   connect((state: StateType, ownProps) => {
     const route = state.cityContent.eventsRouteMapping[ownProps.navigation.getParam('key')]
-    return {invalidLanguage: route && !route.allAvailableLanguages.has(state.cityContent.language || '')}
+    console.log(state)
+    console.log(route)
```
Schedulaar: Remove console.log

PR #58, opened 2019-05-14
NATIVE-154: Copy search component

klinzo:
Should we display a search hint? I think it's pretty self explanatory as it is.


src/modules/endpoint/reducers/pushCategory.js
```diff
@@ -52,7 +52,8 @@ const pushCategory = (state: CityContentStateType, action: PushCategoryActionTyp
         allAvailableLanguages: getAllAvailableLanguages(root, language, city, languages)
       }
     },
-    resourceCache: {...state.resourceCache, ...resourceCache}
+    resourceCache: {...state.resourceCache, ...resourceCache},
+    categoriesMap
```
Schedulaar: Hmm... not sure if this is the right spot to add the categoriesMap to the store... (Because we also have pushEvent and maybe multiple more)... I think I'd rather like to have a separate action for that, which is only called when opening the SearchModal which stores that categoriesMap in sth like `cityContent.searchRoute.categoriesMap` to have some encapsulated space just for the searchRoute 🤔 We could also clean this again once the route has disconnected... This would also follow the prior structure of the state a little bit better...

But you can decide :D

src/routes/search/components/SearchActiveHeader.js
```diff
@@ -0,0 +1,60 @@
...
+class SearchActiveHeader extends React.Component<PropsType> {
+  render () {
+    const {theme, query, closeSearchBar, onSearchChanged} = this.props
+
+    return (
+      <>
+        <BoxShadow theme={theme}>
+          <HorizontalLeft theme={theme}>
+            <HeaderBackButton theme={theme} onPress={closeSearchBar} />
+            <ThemedSearchBar theme={theme} onChangeText={onSearchChanged} value={query} />
```
Schedulaar:
```suggestion
            <ThemedSearchBar theme={theme} onChangeText={onSearchChanged} value={query} ref={search => this.search = search} />
```

src/routes/search/components/SearchActiveHeader.js
```diff
@@ -0,0 +1,60 @@
...
+
+type PropsType = {|
+  theme: ThemeType,
+  query: string,
+  closeSearchBar: () => void,
+  onSearchChanged: (query: string) => void
+|}
+
+class SearchActiveHeader extends React.Component<PropsType> {
+  render () {
```
Schedulaar: Could you try these changes for enabling auto-focus?

(according to https://react-native-training.github.io/react-native-elements/docs/searchbar.html#a-name-calling-a-calling-methods-on-searchbar this should work)

```suggestion
  componentDidMount () {
    this.search.focus()
  }

  render () {
```

src/routes/search/components/SearchActiveHeader.js
```diff
@@ -0,0 +1,60 @@
+// @flow
```
Schedulaar: Maybe rather `SearchHeader`? What does the `active` part mean?

klinzo: Sounds good. Active was meant to signalize that there is an active search and not just a search icon.

src/routes/search/components/SearchActiveHeader.js
klinzo: In fact I tried, but it didn't work, so I created a task. So I would do it in a separate task.

src/modules/endpoint/reducers/pushCategory.js
klinzo: We could move the data to a searchRoute prop, I have absolutely no problem with that. But adding a separate action adds unnecessary complexity imo. An additional saga/reducer would be needed. If at some point also events could be searched, I'd agree with you though. What do you think?

klinzo:
@maxammann no I didn't check how it looks on ios, I don't know how to do that. I didn't change any design stuff though.

src/modules/endpoint/reducers/pushCategory.js
klinzo: @Schedulaar

Schedulaar: Well, I still think, this would be the better and more sound approach. I'd go for it, but I approved  this PR, so you could also merge it now :)

klinzo: Alright, I'll just gonna add a note to the corresponding task.

maxammann:
I check it on iOS and it is working. But there is a minor glitch which is probably caused because it is a modal. The moment you open the modal the Header of the screen below the modal gets removed.
So the animation is not as smooth as it could be because the background is changing.

But I'll create a separate task for this: https://issues.integreat-app.de/browse/NATIVE-224

Feel free to merge!

PR #59, opened 2019-05-17
NATIVE-46: Add pull to refresh

Schedulaar:


src/modules/common/components/NavigationTile.js
```diff
@@ -17,7 +17,7 @@ type PropsType = {|
 |}
 
 const Circle = styled(View)`
-  margin-top: 10px;
+  margin-top: 9px;
```
NKlug: :heart:

src/routes/categories/components/CategoriesScrollView.js
```diff
@@ -0,0 +1,56 @@
...
+    return (
+      <ScrollView refreshControl={<RefreshControl onRefresh={this.onRefresh} refreshing={loading} />}>
+        {!loading && // $FlowFixMe Flow doesn't recognize stateView and cities to not be nullish here -.-
+        <Categories stateView={stateView}
+                    cities={cities}
+                    resourceCache={resourceCache}
+                    language={this.props.language}
+                    cityCode={this.props.cityCode}
+                    theme={this.props.theme}
+                    navigateToCategory={this.props.navigateToCategory}
```
NKlug: Extract everything in line 36 for consistency.

src/modules/endpoint/reducers/cityContentReducer.js
```diff
@@ -17,10 +17,18 @@ export default (
       return pushEvent(state, action)
     case 'MORPH_CONTENT_LANGUAGE':
       return morphContentLanguage(state, action)
-    case 'CLEAR_CATEGORY':
+    case 'FETCH_EVENT':
+    case 'CLEAR_EVENT': {
+      const {key} = action.params
+      delete state.eventsRouteMapping[key]
+      return state
+    }
```
klinzo: What's the use of the brackets? And why would you want to delete stuff on `FETCH_EVENT`/`FETCH_CATEGORY`?

src/routes/landing/containers/LandingContainer.js
```diff
@@ -24,7 +25,28 @@ const mapDispatchToProps = (dispatch: Dispatch<StoreActionType>, ownProps) => {
       type: 'FETCH_CITIES',
       params: {}
     }),
-    navigateToCategory: createNavigateToCategory('Dashboard', dispatch, ownProps.navigation)
+    navigateToDashboard: (cityCode: string, language: string) => {
```
klinzo: Why did you have to change this? Is it because of the key and the problem I also have in #57 ?

klinzo: Can't be maybe just create a separate helper function `createNavigateToDashboard`? Mustn't belong to this task though.

src/modules/endpoint/reducers/cityContentReducer.js
Schedulaar: the brackets are used to scope the variables (case doesn't do that) and therefor I can use `key` as a new const variable in the next case.
I delete the requested route in the FETCH_EVENT/FETCH_CATEGORY to let the screen know, that sth for their screen is being loaded (only necessary if there has been something on the screen which was invalidated by the pull-to-refresh user interaction). (In the future we could also add some `loading` property to the route, but I guess I'd do that in a separate task)

src/routes/landing/containers/LandingContainer.js
Schedulaar: Yes, it was because of the key problem and
Yes, we could create a separate helper function, but we need a specific one to go *from* Landing *to* Dashboard (because we need to address the `switch`-router here as well. So this would be the only occurence here...

klinzo: So I think this should be fine as is.

src/routes/dashboard/components/Dashboard.js
```diff
@@ -80,18 +81,22 @@ class Dashboard extends React.Component<PropsType> {
...
-    }
+    const loading = !stateView || !cities || !resourceCache
 
-    return (<ScrollView>
+    return <ScrollView
+      refreshControl={<RefreshControl onRefresh={this.onRefresh} refreshing={loading} />}>
+      {!loading && <>
         <NavigationTiles tiles={this.getNavigationTileModels()}
                          theme={theme} />
+        {/* $FlowFixMe Flow doesn't recognize stateView and cities to not be nullish */}
```
maxammann: See above.

package.json
```diff
@@ -53,6 +53,7 @@
     "react-native-webview": "^5.3.0",
     "react-navigation": "^3.3.2",
     "react-navigation-header-buttons": "^2.1.1",
+    "react-navigation-stack": "^1.4.0",
```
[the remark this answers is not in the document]
Schedulaar: We have a direct import on that package, so we depend on it directly... I upgraded `react-navigation` as well. https://github.com/react-navigation/react-navigation/blob/v3.10.2/package.json#L42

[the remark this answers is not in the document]
maxammann: Oh oke true yeah we should depend on it :/
As far as I can see react-navigation is still "3.3.2"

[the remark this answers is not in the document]
Schedulaar: Not in the yarn.lock... Should I adjust the package.json as well?

PR #60, opened 2019-05-17
NATIVE-94: Cleanup Resources

Schedulaar:
+13 lines for cleaning up resources :D


src/modules/endpoint/DataContainer.js
```diff
@@ -66,19 +66,13 @@ export interface DataContainer {
    * Returns the ResourceCache.
    * @throws Will throw an error if the ResourceCache is null.
    */
-  getResourceCache: () => Promise<ResourceCacheStateType>,
+  getResourceCache: () => Promise<LanguageResourceCacheStateType>,
 
   /**
-   * Sets the ResourceCache and persists it.
+   * Sets the cache entries for the current city-language-pair and cleans up unnecessary files.
```
NKlug: 'deletes' instead of 'cleans up' might be a little bit more explicit.

maxammann: I think it would be good to split this into 2 function. So setResourceCache stays the same and we add a function like: `cleanupResourceCache`

Usually you do not expect a setter to do much.

Schedulaar: Hmm... I think it's a good encapsulation and prevents misuse of this functionality. (So the resourceCache always stays in sync with the actually persisted files)

Schedulaar: Also: Isn't it a misconception, that setters should not contain logic? 🤔

PR #62, opened 2019-05-21
NATIVE-96: Add errors to state

klinzo:
Not sure if this is the right approach. Please take a brief look and tell me what you think. Also, I don't know where to put errors occurring in `fetchResources`, maybe the `cityContent` state would be the right place for this, but then we had to check three places for errors all the time. Maybe I'm also doing this way to complicated... ^^

https://flow.org/en/docs/types/unions/#toc-disjoint-unions-with-exact-types


src/modules/app/StateType.js
```diff
@@ -24,6 +25,7 @@ export type CategoryRouteStateType = {|
 |}
 
 export type EventRouteStateType = {|
+  +success: true,
```
maxammann: I think this is not needed because you can if there is no `error` then it is a success.

src/modules/app/StoreActionType.js
```diff
@@ -51,7 +48,11 @@ export type FetchCategoryActionType = {|
   |}
 |}
 export type FetchCategoryFailedActionType = {|
-  type: 'FETCH_CATEGORY_FAILED', message: string
+  type: 'FETCH_CATEGORY_FAILED',
+  params: {|
```
maxammann: I'm not sure here whether we should use `params`. The flux spec says to use payload:
https://github.com/redux-utilities/flux-standard-action

I think we can just add them directly to the root of the action.

src/modules/common/selectors/availableLanguagesSelector.js
```diff
@@ -9,11 +9,15 @@ import type {
 const cityLanguagesSelector = (state: StateType): ?Array<string> => state.cityContent.languages &&
   state.cityContent.languages.map(languageModel => languageModel.code)
 
-const categoryRouteSelector = (state: StateType, ownProps): ?CategoryRouteStateType =>
-  state.cityContent.categoriesRouteMapping[ownProps.navigation.getParam('key')]
+const categoryRouteSelector = (state: StateType, ownProps): ?CategoryRouteStateType => {
+  const route = state.cityContent.categoriesRouteMapping[ownProps.navigation.getParam('key')]
+  return route.success && route
```
maxammann: I could be wrong, but there is no gurantee that sucess does exist. So this should throw a flow error.

src/modules/endpoint/reducers/__tests__/morphContentLanguage.spec.js
```diff
@@ -382,6 +382,7 @@ describe('morphContentLanguage', () => {
 
     const previous = prepareState()
 
+    // $FlowFixMe
     previous.categoriesRouteMapping['route-0'].models['/augsburg/de/anlaufstellen'] = undefined
```
maxammann: Not sure why this FlowFixMe is need. Add a comment here.

src/modules/app/StateType.js
```diff
@@ -37,6 +39,11 @@ export type FileCacheStateType = {
   |}
 }
 
+export type ErrorStateType = {|
+  +error: true,
+  +message: string
```
maxammann: You can change this to: 
```
errorMessage: ?string
```

maxammann: or just `errorMessage` because if it is null then there should be a full `CategoryRouteStateType`

src/modules/app/StateType.js
```diff
@@ -37,6 +39,11 @@ export type FileCacheStateType = {
   |}
 }
 
+export type ErrorStateType = {|
+  +error: true,
```
maxammann: If this is a boolean then this should be called something like `loadingFailed` `errorHappened`

src/modules/app/StateType.js
klinzo: Yes, it is required. See: https://flow.org/en/docs/types/unions/#toc-disjoint-unions-with-exact-types

src/modules/app/StoreActionType.js
klinzo: AFAIK we use params nearly all the time. So I changed this for the sake of consistency. If we don't want this, we should do this in a task imo.

src/modules/common/selectors/availableLanguagesSelector.js
klinzo: Why should this throw a flow error? If `route.success` does not exist, `route.success` is `undefined` and so it returns `undefined`, which is okay since the return type is `?CategoryRouteState?

src/modules/app/StateType.js
klinzo: Nope, not possible. See https://flow.org/en/docs/types/unions/#toc-disjoint-unions-with-exact-types . A full `CategoryRouteStateType` AND an error makes no sense. So if there is no `CategoryRouteStateType`, there has to be an `ErrorStateType` and vice versa. This should really be an XOR

src/modules/app/StateType.js
klinzo: I just used the flow terminology

klinzo:
@maxammann The description of the `wip` label is `DO NOT REVIEW YET`, but I would like reviews.

maxammann:
You attached needs-review label 3 days ago

maxammann:
Thats my review/feedback

src/modules/app/StoreActionType.js
maxammann: If it is more consistent now then this is good :+1:

src/modules/common/selectors/availableLanguagesSelector.js
maxammann: I thought that something like that does not work but it does:
```
type AaType = {|
  a: boolean
|}

type BbType = {|
  b: boolean
|}

type CcType = AaType | BbType

const c: CcType = {b: true}

if (c.a) {
  console.log()
}
```

So this is like a type check.

src/modules/app/StateType.js
maxammann: Yes this is a  "either/or" relationship. But this structure contains redundancy if I get it right.
Like: If message is not null or undefined then error is always true.

src/modules/app/StateType.js
maxammann: Which flow terminology?

klinzo: The one from  the flow docs, https://flow.org/en/docs/types/unions/#toc-disjoint-unions-with-exact-types :D

src/modules/app/StateType.js
klinzo: Yeah I know, but this is the recommended way of flow to do smth like this. If this redundancy is removed,  flow throws errors, because flow doesn't get it anymore. E.g. if you have types `type ErrorType = {|
  error:  string
|}` and `type SuccessType = {|
   data: XY,
  moreData: UV
|}` and `type SomeType = ErrorType | SuccessType` and you check for error like this:
`if (someObject.error) {
  do something
} else {
  do something with someObject.data
}`,
flow throws an error that someObject.data is not defined. You'd always have to check for all props you want to use,  e.g. `if (someObject.data && someObject.moreData) { do something with data and moreData }`

klinzo:
@Schedulaar your opinion on this?

src/modules/app/StateType.js
maxammann: Hm if I understand you correctly something like this should not work, right?

```
type AaType = {|
  errorMessage: string
|}

type BbType = {|
  result: any
|}

type CcType = AaType | BbType

const c: CcType = {result: 'true'}

if (c.errorMessage) {
  console.log(c.errorMessage)
  const r: AaType = c
  console.log(r)
}

if (c.result) {
  console.log(c.result)
  const r: BbType = c
  console.log(r)
}
```

klinzo: Okay, we have to check for `undefined` to make it working.

src/modules/endpoint/reducers/__tests__/morphContentLanguage.spec.js
```diff
@@ -393,6 +393,9 @@ describe('morphContentLanguage', () => {
 
     const previous = prepareState()
 
+    if (previous.categoriesRouteMapping.errorMessage !== undefined) {
+      throw Error('Preparation of state failed')
```
maxammann: I think you can add the errorMessage to the Error here

src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -4,59 +4,63 @@ import { withTheme } from 'styled-components/native'
...
-    }
+  const {resourceCache, categoriesRouteMapping, languages, language, city} = state.cityContent
+
+  if (languages && !languages.map(languageModel => languageModel.code).includes(language)) {
+    return {invalidLanguage: true}
+  }
+
+  if (state.cities.errorMessage !== undefined ||
+    categoriesRouteMapping.errorMessage !== undefined ||
+    resourceCache.errorMessage !== undefined) {
```
maxammann: Alternative for check:

```
if ([state.cities.errorMessage, resourceCache.errorMessage, categoriesRouteMapping.errorMessage].includes(undefined)) {
```

klinzo: For some reasons both suggestions are producing new flow errors. I'll just merge it.

PR #63, opened 2019-05-24
NATIVE-212: Support post requests for extras

Schedulaar:


index.js
```diff
@@ -1,3 +1,4 @@
+import 'core-js'
```
maxammann: Is there a reason we really need that? I required fromEntries aswell today but used reduce instead.

Shouldn't the bable depedency take care of all polyfills?

package.json
```diff
@@ -67,27 +69,27 @@
...
+    "babel-plugin-module-resolver": "^3.2.0",
     "babel-plugin-transform-inline-environment-variables": "^0.4.3",
     "babel-plugin-transform-remove-console": "^6.9.4",
-    "eslint": "^5.12.1",
+    "eslint": "^5.16.0",
     "eslint-config-standard": "^12.0.0",
     "eslint-config-standard-react": "^7.0.2",
-    "eslint-plugin-flowtype": "^3.2.1",
-    "eslint-plugin-import": "^2.13.0",
-    "eslint-plugin-jest": "^22.1.3",
```
maxammann: Are these Upgrades compatible with our eslint packages? yarn shows a Warnung in that case.

src/routes/external-extra/components/ExternalExtra.js
```diff
@@ -0,0 +1,38 @@
...
+export type PropsType = {|
+  url: ?string,
+  postData: ?Map<string, string>
+|}
+
+class ExternalExtra extends React.Component<PropsType> {
+  render () {
+    const {url, postData} = this.props
+    // $FlowFixMe Object.fromEntries is available in Flow version >= 0.99.x: https://github.com/facebook/flow/commit/f92231cf15022c729226b4d440435b3366f9346c
+    const body = !postData ? '' : stringify(Object.fromEntries(postData))
```
maxammann: I think we should avoid adding FlowFixMes if there is an easy workaround like a `reduce`. I mean yes it is not pretty, but we will never have time to fix cosmetics like this ^^

src/routes/external-extra/components/ExternalExtra.js
```diff
@@ -0,0 +1,38 @@
...
+    return <WebView
+      source={{
+        uri: url,
+        method: postData ? 'POST' : 'GET',
```
maxammann: You said on Friday that it did not work because of specific Headers. Was that the case or was there an other problem?

src/routes/external-extra/components/ExternalExtra.js
Schedulaar: Well, this FlowFixMe is exceptionally in the manner, that I already know that it's going to be fixed with a newer flow version, so I'd reject your comment from that point of view :) Also lodashs reduce does not work with a map or es6 iterators out of the box... But in fact, it might be possible using the array destructor and lodash's `toPairs`... I'll give it a try...

src/routes/external-extra/components/ExternalExtra.js
Schedulaar: Well the problem was that I sent json in the body together with a `application/x-www-form-urlencoded` Content-Type header. I tried to adjust the Content-Type header, but this is not possible together with a POST request in Android (wtf). So I decided to use `stringify` from `query-string` instead of a JSON body.
This works on Android but not on iOS (wtf). So this PR is still WIP...

package.json
Schedulaar: I didn't see any warnings...

index.js
Schedulaar: It only takes care of some polyfills, I guess... It's not really transparent to me, which ones though... Seems like it doesn't polyfill proposals...

I removed the `core-js` dependency again...

src/routes/external-extra/components/ExternalExtra.js
Schedulaar: Did work 👍

src/routes/external-extra/components/ExternalExtra.js
Schedulaar: Whoops, it *does* work on iOS 🤔 🎉  (maybe because of replacing `Object.fromEntries`).

Schedulaar: Okay, now it doesn't work again iOS (wtf)... I'll have to check this once again. Not sure how to debug this better...

Schedulaar: Now it works with the added `Content-Type` header 🎉  (Proof:
image
)

src/routes/external-extra/containers/ExternalExtraContainer.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+
+import { mapProps } from 'recompose'
+import ExternalExtra from '../components/ExternalExtra'
+import type { NavigationScreenProp } from 'react-navigation'
+import type { PropsType } from '../components/ExternalExtra'
+
+export default mapProps<PropsType, {navigation: NavigationScreenProp<*>}>(
```
klinzo: Make this strict. Do we have a linting rule for strict props?

src/routes/extras/containers/ExtrasContainer.js
```diff
@@ -52,11 +52,14 @@ class ExtrasContainer extends React.Component<PropsType, ExtrasStateType> {
     this.loadExtras()
   }
 
-  navigateToExtra = (path: string, isExternalUrl: boolean, offerHash: ?string = null) => {
+  navigateToExtra = (path: string, isExternalUrl: boolean, postData: ?Map<string, string>) => {
+    if (!this.props.navigation.push) {
+      return console.error('push is not defined on navigation')
```
klinzo: Is there a possibility that `push` is not defined? Or is this just a flow problem?

maxammann: That is a very weird problem with flow types of react-navigation. Couldn't fix it back then.

I would prefer:
```
throw Error('push is not defined on navigation')
```
which crashes the app and reports it though sentry.

package.json
maxammann: You only see them when you remove the `node_modules` dir and rerun flow.

maxammann: If there is no reason for upgrading eslint I would suggest to downgrade them again. I'm not really sure what different version of eslint could cause.

If you feel confident ignore this comment.

src/routes/external-extra/components/ExternalExtra.js
maxammann: Many wtfs :D

src/routes/extras/containers/ExtrasContainer.js
maxammann: Ofc a real fix would be cool :)

PR #64, opened 2019-05-25
NATIVE-11: Add settings screen

maxammann:


src/routes/language/components/ChangeLanguageModal.js
```diff
@@ -19,25 +20,35 @@ const Wrapper = styled.View`
...
     InteractionManager.runAfterInteractions(() => {
       changeLanguage(city, model.code)
     })
   }
 
   render () {
-    const { theme, languages, availableLanguages, currentLanguage } = this.props
+    const {theme, languages, availableLanguages, city, currentLanguage} = this.props
+
+    if (!languages || !currentLanguage || !city) {
```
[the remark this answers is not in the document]
klinzo: Or at least handle this stuff in the containers.

src/routes/search/components/SearchModal.js
```diff
@@ -19,20 +20,23 @@ const Wrapper = styled.View`
...
+  closeModal: () => void,
+  navigation: NavigationScreenProp<*>
 |}
 
 type StateType = {|
   query: string
 |}
 
 class SearchModal extends React.Component<PropsType, StateType> {
+  static navigationOptions = {}
```
klinzo: Why is this required?

src/routes/search/components/SearchModal.js
```diff
@@ -19,20 +20,23 @@ const Wrapper = styled.View`
...
-  theme: ThemeType,
-  language: string,
-  cityCode: string,
-  closeModal: () => void
+  navigateToCategory: (cityCode: string, language: string, path: string) => StoreActionType,
+  theme: any,
+  language: string | null,
+  cityCode: string | null,
+  closeModal: () => void,
+  navigation: NavigationScreenProp<*>
```
klinzo: Why is `navigation` required?

src/routes/search/containers/SearchModalContainer.js
```diff
@@ -7,20 +7,36 @@ import type { StoreActionType } from '../../../modules/app/StoreActionType'
 import { withTheme } from 'styled-components/native'
 import createNavigateToCategory from '../../../modules/app/createNavigateToCategory'
 import SearchModal from '../components/SearchModal'
+import { CategoriesMapModel } from '@integreat-app/integreat-api-client'
+import type { NavigationScreenProp } from 'react-navigation'
 
-const mapStateToProps = (state: StateType, ownProps) => {
+type OwnPropsType = {| navigation: NavigationScreenProp<*> |}
+
+export type PropsType = {|
```
klinzo: Shouldn't this be the PropsTypes from `SearchModal`? At least it's a duplicate.

src/routes/settings/components/SwitchSetting.js
```diff
@@ -0,0 +1,30 @@
...
+import * as React from 'react'
+import { Switch } from 'react-native'
+
+type PropType = {
+  onValueChange: boolean => void,
+  getCurrentValue: () => boolean
+}
+
+type StateType = {
+  value: boolean
```
klinzo: This is a duplication of the state. It should be enough to pass the value as prop.

src/routes/search/containers/SearchModalContainer.js
klinzo: Alright, I see. Many PropTypes now. Maybe rename to ConnectPropTypes. Not sure though.

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,197 @@
...
+      const settingsArray = await AsyncStorage.multiGet(settingsKeys)
+
+      const settings = reduce(settingsArray,
+        (accumulator, [key, value]) => {
+          accumulator[key] = JSON.parse(value)
+          return accumulator
+        },
+        {})
+
+      this.setState(state => ({...state, settingsLoaded: true, settings}))
```
klinzo: You can simplify this to `this.setState({settingsLoaded: true, settings})

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,197 @@
...
+    } catch (e) {
+      console.error('Failed to load settings.')
+    }
+  }
+
+  async setSetting (changeSetting: SettingsType => $Shape<SettingsType>) {
+    this.setState(state => {
+      const newSettings = changeSetting(state.settings)
+      return ({...state, settings: {...state.settings, ...newSettings}})
```
klinzo: Could be simplified to `this.setState({settings: {...changeSetting(state.settings)}})`

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,197 @@
...
+import styled from 'styled-components/native'
+
+import SettingItem from './SettingItem'
+import type { ThemeType } from '../../../modules/theme/constants/theme'
+import type { NavigationScreenProp } from 'react-navigation'
+import { reduce } from 'lodash/collection'
+import type { TFunction } from 'react-i18next'
+
+type PropsType = {
+  navigation: NavigationScreenProp<*>,
```
klinzo: Why is navigation needed?

src/routes/settings/container/SettingsContainer.js
```diff
@@ -0,0 +1,25 @@
...
+import type { NavigationScreenProp } from 'react-navigation'
+import { translate } from 'react-i18next'
+import { connect } from 'react-redux'
+import type { StateType } from '../../../modules/app/StateType'
+
+type OwnPropsType = {| navigation: NavigationScreenProp<*> |}
+
+type PropsType = {
+  navigation: NavigationScreenProp<*>
+}
```
klinzo: Should be strict, and AFAIK it should only contain a `language: string`

src/modules/app/components/Navigator.js
```diff
@@ -60,10 +61,10 @@ export const ExtrasStack = createStackNavigator(
 export const AppStack = createStackNavigator(
   {
     'Dashboard': createNavigationScreen(LayoutedDashboardContainer, defaultHeader),
-    // $FlowFixMe No idea why this fails
```
Schedulaar: ❤️

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,197 @@
...
+              }
+            }
+          },
+          {
+            title: t('privacyPolicy'),
+            onPress: () => {
+              if (language === 'de') {
+                Linking.openURL('https://integreat-app.de/datenschutz-webseite/')
+              } else {
+                Linking.openURL('https://integreat-app.de/en/privacy-website/')
```
Schedulaar: We might want to extract all these urls to some constants file

src/routes/settings/components/SwitchSetting.js
Schedulaar: I agree. The only purpose of this component is to "cache" the value. Since this component is not even used anywhere I guess, it can be removed...

src/routes/settings/container/SettingsContainer.js
```diff
@@ -0,0 +1,25 @@
...
+
+const mapStateToProps = (state: StateType) => {
+  return {
+    language: state.cityContent.language
+  }
+}
+
+const ThemedSettings = withTheme<PropsType, _>(translate('settings')(Settings))
+export default connect<PropsType, OwnPropsType, _, _, _, _>(mapStateToProps)(
```
Schedulaar: Haha, this is super ugly :D but oh well^^

src/routes/settings/components/Settings.js
Schedulaar: Not sure about this one... I think React only shallowly merges the state. So you could simplify it to
```js
this.setState(state => {settings: {...state.settings, ...changeSetting(state.settings)}})
```

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,197 @@
...
+        extraData={this.state.settings}
+        renderItem={this.renderItem}
+        renderSectionHeader={this.renderSectionHeader}
+        ItemSeparatorComponent={this.ThemedItemSeparator}
+        SectionSeparatorComponent={this.ThemedItemSeparator}
+        stickySectionHeadersEnabled={false}
+      />
+    )
+  }
+}
```
Schedulaar: Can you reduce the size of this file?

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,197 @@
...
+      const newSettings = changeSetting(state.settings)
+      return ({...state, settings: {...state.settings, ...newSettings}})
+    })
+
+    const settingsArray = reduce(changeSetting(this.state.settings),
+      (accumulator, value, key) => {
+        accumulator.push([key, JSON.stringify(value)])
+        return accumulator
+      },
+      [])
```
Schedulaar: With lodash you can do 
```js
const settingsArray = toPairs(mapValues(changeSetting(this.state.settings), value => JSON.stringify(value)))
```
Also you should probably do the AsyncStorage stuff in the callback of `setState`, right? Or is it not possible with promises?
What do you think of https://medium.com/front-end-weekly/async-await-with-react-lifecycle-methods-802e7760d802

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,197 @@
...
+
+  constructor (props: PropsType) {
+    super(props)
+
+    this.state = {settingsLoaded: false, settings: {errorTracking: false, test: false}}
+
+    this.loadSettings()
+  }
+
+  async loadSettings () {
```
Schedulaar: Probably this should not be part of the settingsScreen, because we would want to load Settings  from other parts of the app as well?

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,197 @@
...
+
+type SectionType = { title: ?string }
+
+const ItemSeparator = styled.View`
+    background-color: ${props => props.theme.colors.textDecorationColor};;
+    height: ${StyleSheet.hairlineWidth};
+`
+
+const SectionHeaderContainer = styled.View`
+`
```
Schedulaar: remove if not needed.

src/routes/settings/components/Settings.js
```diff
@@ -0,0 +1,197 @@
...
+            onPress: () => {
+              if (language === 'de') {
+                Linking.openURL('https://integreat-app.de/datenschutz-webseite/')
+              } else {
+                Linking.openURL('https://integreat-app.de/en/privacy-website/')
+              }
+            }
+          },
+          {
+            title: t('version', {version: '??'})
```
Schedulaar: Add version or todo comment + task

src/modules/layout/components/Header.js
```diff
@@ -103,7 +107,8 @@ class Header extends React.PureComponent<PropsType> {
   }
 
   render () {
-    const { theme } = this.props
+    const { theme } = this.props // TODO: When mering the share button task make sure to remove the mock t function
+    const t = (key: string) => key
```
Schedulaar: *reminder*

src/routes/language/components/ChangeLanguageModal.js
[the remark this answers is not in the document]
maxammann: We can create a task for this, as this only came up because of fixing the flow types.

But unfortunately that task will not have priority. Probably a hoc would make sense, but only if and only if the flow types work good.

The benefit of doing it in the render method is that it is very simple.

src/routes/search/components/SearchModal.js
maxammann: Just added it for consistency

src/routes/settings/components/SwitchSetting.js
maxammann: Yes this is a left-over. Thanks for pointing out!

src/routes/settings/components/Settings.js
maxammann: Both do not work because they would overwrite `settingsLoaded`

You can only set the whole state with `setState`

src/routes/settings/components/Settings.js
maxammann: Settings will get the navigation prop because connect and withTheme passes it through. There are still some problems with the flow types but it works oke for now.

src/routes/settings/components/Settings.js
maxammann: We could not as these are not reused I see not really a point in doing it.

src/routes/settings/container/SettingsContainer.js
maxammann: Yes it is.. ... complicated

src/modules/layout/components/Header.js
maxammann: You were faster! Pfff!

maxammann: Didn't even review your changes to the share task! :D

src/routes/settings/components/Settings.js
maxammann: Hm not sure yet, but I think I'll do this in the sentry task.

src/modules/app/components/Navigator.js
maxammann: Took me hours :D

src/routes/search/components/SearchModal.js
klinzo: Consistency with what? Is this common?

src/routes/settings/components/Settings.js
Schedulaar: Have you tried? Usually setState merges shallowly!

src/routes/search/components/SearchModal.js
```diff
@@ -19,20 +20,23 @@ const Wrapper = styled.View`
...
-type PropsType = {|
+export type PropsType = {|
   categories: CategoriesMapModel | null,
-  navigateToCategory: (cityCode: string, language: string, path: string) => void,
-  theme: ThemeType,
-  language: string,
-  cityCode: string,
-  closeModal: () => void
+  navigateToCategory: (cityCode: string, language: string, path: string) => StoreActionType,
+  theme: any,
```
klinzo: Why did you change this to any? Problems with flow?

src/routes/search/components/SearchModal.js
```diff
@@ -19,20 +20,23 @@ const Wrapper = styled.View`
...
 
-type PropsType = {|
+export type PropsType = {|
   categories: CategoriesMapModel | null,
-  navigateToCategory: (cityCode: string, language: string, path: string) => void,
-  theme: ThemeType,
-  language: string,
-  cityCode: string,
-  closeModal: () => void
+  navigateToCategory: (cityCode: string, language: string, path: string) => StoreActionType,
```
klinzo: There should be a more specific type for this, shouldn't it?

src/routes/search/components/SearchModal.js
klinzo: I mean e.g. in the `Dashboard` component you removed the `navigationOptions` completely.

src/routes/search/components/SearchModal.js
```diff
@@ -61,8 +65,13 @@ class SearchModal extends React.Component<PropsType, StateType> {
       .map(category => ({model: category, subCategories: []}))
   }
 
-  onItemPress = (category: {path: string}) => {
+  onItemPress = (category: { path: string }) => {
```
klinzo: Should we try to always use strict props again?

src/routes/language/components/ChangeLanguageModal.js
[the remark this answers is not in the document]
maxammann: https://issues.integreat-app.de/browse/NATIVE-53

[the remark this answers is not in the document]
maxammann: Old but gold task :D

src/routes/search/components/SearchModal.js
maxammann: We could force this with a linting rule. But there are a lot of places right now where it could be used. 

I think it makes most sense for PropTypes and the state. If we have them there I'm fine.

src/routes/settings/components/Settings.js
maxammann: _Note if you do return a promise from the lifecycle methods it will be not be awaited._

So you can return it but it will not make a difference

src/routes/search/components/SearchModal.js
maxammann: I tried to use the specific type but it does not work with navigateToCategory right now

PR #65, opened 2019-05-28
NATIVE-202: Write to a temp location before moving it to the target

maxammann:


maxammann:
Yes, this is already done on iOS

PR #66, opened 2019-05-28
NATIVE-188: Always run fetcher

maxammann:


src/modules/endpoint/sagas/loadCategories.js
```diff
@@ -17,37 +17,44 @@ function * fetchCategoriesMap (city: string, language: string): Saga<CategoriesM
...
+    console.debug('Fetching categories')
 
-  console.debug('Fetching categories')
+    // TODO: data was loaded but should be incrementally updated. This will be done in NATIVE-3
 
-  // TODO: data was loaded but should be incrementally updated. This will be done in NATIVE-3
+    const categoriesMap: ?CategoriesMapModel = yield call(fetchCategoriesMap, city, language)
 
-  const categoriesMap: ?CategoriesMapModel = yield call(fetchCategoriesMap, city, language)
+    if (!categoriesMap) {
```
klinzo: Is it possible that `fetchCategoriesMap` returns null or undefined? As far as I can see, it's not possible (or the flow types in `fetchCaegoriesMap` are wrong. So imo this check could be removed.

src/modules/endpoint/sagas/loadEvents.js
```diff
@@ -23,19 +23,25 @@ function * fetchEvents (city: string, language: string): Saga<?Array<EventModel>
...
+    console.debug('Fetching events')
 
-  console.debug('Fetching events')
+    // TODO: if data was loaded but should be updated incrementally. This will be done in NATIVE-3
 
-  // TODO: if data was loaded but should be updated incrementally. This will be done in NATIVE-3
+    events = yield call(fetchEvents, city, language)
 
-  const events: ?Array<EventModel> = yield call(fetchEvents, city, language)
+    if (!events) {
```
klinzo: Same as above. I think the types are wrong here also in `fetchEvents`

src/modules/endpoint/sagas/loadCategories.js
```diff
@@ -17,37 +17,44 @@ function * fetchCategoriesMap (city: string, language: string): Saga<CategoriesM
...
+      throw new Error('Failed to load categories!')
+    }
 
-  if (!categoriesMap) {
-    throw new Error('Failed to load categories!')
-  }
+    yield call(dataContainer.setCategoriesMap, categoriesMap)
 
-  const categories = categoriesMap.toArray()
+    categories = categoriesMap
```
klinzo: You could use `categories` instead of using a separate `categoriesMap` var from the start.

src/modules/endpoint/sagas/loadCategories.js
```diff
@@ -17,37 +17,44 @@ function * fetchCategoriesMap (city: string, language: string): Saga<CategoriesM
...
+    console.debug('Using cached categories')
+    categories = yield call(dataContainer.getCategoriesMap)
+  }
 
   const resourceURLFinder = new ResourceURLFinder()
   resourceURLFinder.init()
 
   const urls = resourceURLFinder.buildFetchMap(
-    categories,
+    categories.toArray(),
```
klinzo: Did that work before? :O

src/modules/endpoint/sagas/loadEvents.js
```diff
@@ -23,19 +23,25 @@ function * fetchEvents (city: string, language: string): Saga<?Array<EventModel>
 
 function * loadEvents (
   city: string, language: string, dataContainer: DataContainer, shouldUpdate: boolean): Saga<FetchMapType> {
-  if (dataContainer.eventsAvailable() && !shouldUpdate) {
-    console.debug('Using cached events')
-    return {}
-  }
+  let events: ?Array<EventModel>
+
+  // If data was loaded and should not be updated, return
```
Schedulaar: Where's this return that you announce here? :D

src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -35,9 +35,13 @@ export default function * loadCityContent (
     call(loadLanguages, newCity, dataContainer, shouldUpdate)
   ])
 
+  // fetchResourceCache should always be called. Even if loadCategories, loadEvents, loadLanguages did not update the
+  // dataContainer this is needed. In case the previous call to fetchResourceCache failed to download some resources an
+  // other call could fix this and download missing files.
```
Schedulaar: As discussed in the comments of the issue, we would only want to fetchResourceCache if we update all other contents?! This makes updating content a lot simpler and it was the case before this task.

src/modules/endpoint/sagas/loadEvents.js
```diff
@@ -23,19 +23,25 @@ function * fetchEvents (city: string, language: string): Saga<?Array<EventModel>
 
 function * loadEvents (
   city: string, language: string, dataContainer: DataContainer, shouldUpdate: boolean): Saga<FetchMapType> {
-  if (dataContainer.eventsAvailable() && !shouldUpdate) {
-    console.debug('Using cached events')
-    return {}
-  }
+  let events: ?Array<EventModel>
```
Schedulaar: You could also have `const events = yield call(dataContainer.getEvents)` after the first if block (so we don't have `let`)

src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -35,9 +35,13 @@ export default function * loadCityContent (
...
+  const fetchMap = {...categoryUrls, ...eventUrls}
+  yield call(fetchResourceCache, newCity, newLanguage, fetchMap, dataContainer)
+
   if (shouldUpdate) {
-    const fetchMap = {...categoryUrls, ...eventUrls}
-    yield call(fetchResourceCache, newCity, newLanguage, fetchMap, dataContainer)
```
Schedulaar: Here we called `fetchResourceCache` every time we actually did a network request to update jsons.

src/modules/endpoint/sagas/loadCityContent.js
maxammann: Yes the issue description is invalid. I'll update it.
We have to call it each time you open a city because of the reasons I mention in the comment.

src/modules/endpoint/sagas/loadEvents.js
maxammann: Seems like a Copy&Pasta mistake :D

maxammann:
Yes seems like the check whether it is the first time a "FETCH" is triggered is missing. (e.g. when opening a city)

maxammann:
@Schedulaar @klinzo do you have an opinion on this task? If not I'll fix the few points.

maxammann:
@Schedulaar @klinzo do you have an opinion on this task? If not I'll fix the few points.

klinzo:
@maxammann I think the issue makes sense, but  it should only fetch when opening a city and not every time the user is navigating.

Schedulaar:
I think it makes our content-fetching process a lot more complex. Imo it is enough if we fetch the resources only when we fetch the jsons: Especially once incremental refresh is established the jsons will be very small and won't have any weight...

maxammann:
@Schedulaar this is actually the simplest way (with the least complexity), which solves the problem of retrying failed downloads.

Incremental refresh is not a solution to this problem.

Also the cache is handles properly this way.

src/modules/endpoint/sagas/loadEvents.js
maxammann: Yes but this could take longer because work could possibly be done twice.

E.g. when getEvents does a sql query.

src/modules/endpoint/sagas/loadCityContent.js
maxammann: As you did not comment on this one I take this as an: "OK"

src/modules/endpoint/sagas/loadCategories.js
maxammann: It was done in an other line

src/modules/endpoint/sagas/loadCategories.js
maxammann: I removed the `?`. Sags is a bit weak on flow types...

maxammann:
@Schedulaar yes in that case we need to dispatch the correct fetch with `shouldRefreshResources`

PR #67, opened 2019-05-28
NATIVE-229: Fix moment

maxammann:


src/modules/app/components/App.js
```diff
@@ -3,7 +3,7 @@
 import * as React from 'react'
 
 import { Provider } from 'react-redux'
-import I18nProvider from 'modules/i18n/containers/I18nProvider'
+import I18nProvider from '../../../modules/i18n/containers/I18nProvider'
 import createReduxStore from '../createReduxStore'
 import CustomThemeProvider from '../../theme/containers/CustomThemeProvider'
 import IOSSafeAreaView from 'modules/platform/components/IOSSafeAreaView'
```
Schedulaar: Here's an absolute path as well 🤔

src/modules/i18n/containers/I18nProvider.js
```diff
@@ -74,26 +74,26 @@ export class I18nProvider extends React.Component<PropsType, {
...
-  initLanguage (language: string) {
-    const targetLanguage = this.getPredeterminedLanguage(language)
+  initLanguage () {
+    const targetLanguage = this.getI18nextLanguage()
 
     const fonts = I18nProvider.getSelectedFonts(targetLanguage)
     this.setState({language: targetLanguage, fonts})
     this.props.setUiDirection(RTL_LANGUAGES.includes(targetLanguage) ? 'rtl' : 'ltr')
+
+    moment.locale(targetLanguage)
```
Schedulaar: kind of bad style if we set this globally, isn't it?

src/modules/app/components/App.js
maxammann: Yeah there is a task for that :D

src/modules/i18n/containers/I18nProvider.js
maxammann: I think this is fine because we somehow want to tell moment what language it should use by default. Correct me if I'm wrong.

moment already maintains a global language which is usually "en_US".

Yes it is bad design by moment. But I'm not sure whether we can handle it it a different was properly.
Formatting can happen in libraries which use moment. If we want them to format it for the current language we have to do it like this.

maxammann: @klinzo  any opinion on this? 

I'm thinking about removing this line and statically providing a function to create a localized moment date. Something like: `this.context.localizeDate(moment(...)).format()`

klinzo: I'm okay with setting it globally. :man_shrugging: If we don't want it, we should create an issue and deal with it later. IMO this is not something we should invest much time for right now, there are more important issues.

klinzo: Are there any potential problems setting it globally?

maxammann: No, just bad style. Which actually is a problem :D

Also it is not testable.

klinzo: Okay, than you should probably do it if it doesn't take much time and create an issue else.

maxammann: I fixed it by introducing an new context. I also fixed a lot of flow errors.

Schedulaar: So you can remove it?

src/routes/landing/containers/LandingContainer.js
```diff
@@ -59,11 +81,10 @@ const mapDispatchToProps = (dispatch: Dispatch<StoreActionType>, ownProps) => {
 }
 
 export default compose([
-  // withI18n has to be before connect, because we need to pass the language as prop
-  withI18n(),
-  connect(mapStateToProps, mapDispatchToProps),
+  // translate has to be before connect, because we need to pass the language as prop
```
Schedulaar: I don't understand this comment anyway 🙈

src/modules/i18n/containers/I18nProvider.js
maxammann: upsi forgot that :)

PR #68, opened 2019-05-28
NATIVE-246: Fix deadlock on master node

maxammann:
When more jobs are started than there are executors at the same time no job can finish.

This is because currently an other executor is needed to run the compilation. One is used as supervisor.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ void uploadToBrowserstack(String path) {
 }
 
 pipeline {
-    agent any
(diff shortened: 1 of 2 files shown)


maxammann:
@Schedulaar @klinzo I merged this into all open PRs!

klinzo:
So if I merge my PR now, I also move yours? :D Is that okay?

maxammann:
Yes, so lets wait for @Schedulaar s review

PR #69, opened 2019-05-30
NATIVE-112: Display errors

klinzo:


src/modules/common/hocs/withLanguageNotAvailable.js
```diff
@@ -0,0 +1,35 @@
...
+import LanguageNotAvailableContainer from '../containers/LanguageNotAvailableContainer'
+
+export type LanguageNotAvailableType = {|
+  languageNotAvailable: boolean,
+  city: string,
+  languages: Array<LanguageModel>,
+  changeUnavailableLanguage: (city: string, newLanguage: string) => void
+|}
+
+type PropsType<T> = LanguageNotAvailableType | T
```
maxammann: As T is only `*` this generic is unused. Also  think you are handling the tyoe of PropsType wrong in these HOCs.

There should be examples in the webapp or in some other HOC.

src/modules/error/components/Failure.js
```diff
@@ -19,19 +18,17 @@ margin-bottom: 10px;
...
   t: TFunction
 |}
 
 export class Failure extends React.Component<PropsType> {
   render () {
-    const { t } = this.props
+    const { t, error } = this.props
 
     return <ViewContainer>
       <IconContainer name='sentiment-dissatisfied' size={55} color={'black'} />
```
maxammann: Replace that icon with this one: https://drive.google.com/drive/u/1/folders/1CUb_SWII1WhLzbm2_z99WtCqTXpdRHrO

I think it makes sense to do it in this task.

src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -56,11 +52,20 @@ const mapStateToProps = (state: StateType, ownProps) => {
...
 export default compose([
-  withRouteCleaner,
   connect(mapStateToProps, mapDispatchToProps),
-  // TODO NATIVE-112 Show errors
-  branch(props => props.error, renderComponent(Failure)),
-  branch(props => props.invalidLanguage, renderComponent(CategoryLanguageNotAvailableContainer)),
+  withRouteCleaner,
+  withError,
+  withLanguageNotAvailable,
+  withLoading,
```
maxammann: I'm not sure whether this is getting a bit out of hands with hocs. I don't have a better idea right now.
Maybe we can make withError and withLoading a single HOC.

klinzo: Yep, we could just use a single HOC for loading, errors and language not available, which is also kind of an error.

src/routes/dashboard/containers/DashboardContainer.js
```diff
@@ -54,12 +56,36 @@ const mapStateToProps = (state: StateType, ownProps) => {
...
 export default compose([
-  withRouteCleaner,
   connect(mapStateToProps, mapDispatchToProps),
-  // TODO NATIVE-112 Show errors, maybe use withError and pass more props
-  branch(props => props.error, renderComponent(Failure)),
-  branch(props => props.invalidLanguage, renderComponent(LanguageNotAvailableContainer)),
+  withRouteCleaner,
+  withError,
+  withLanguageNotAvailable,
+  withLoading,
```
Schedulaar: not sure about these `withLoading`s as well... As of now this is handled in the `Dashboard` directly using the ScrollView in the RefreshControl. This allows the pull-to-refresh animation not to be interrupted by changing the view to the ActivityIndicator.
If you want to keep them, you would still need to touch Dashboard.js Events.js CategoriesScrollView.js and remove the loading-logic there.

maxammann: Apparently hooks are solving issues like this. The library recompose was abandoned because of hooks.

So maybe we can let it stay like this and refactor this in the future.

Schedulaar: I guess, my comment is still valid and needs to be resolved though...
Either remove the `withLanding`s again or remove the RefreshControl loading-logic from the routes

klinzo: Yeah right. Will do that

klinzo:
@maxammann Updated flow types and removed withLoading again.

src/modules/common/containers/LanguageNotAvailableContainer.js
```diff
@@ -1,51 +1,11 @@
...
-    changeLanguage: (city: string, newLanguage: string) => {
-      const switchContentLanguage: SwitchContentLanguageActionType = {
-        type: 'SWITCH_CONTENT_LANGUAGE',
-        params: {
-          city,
-          newLanguage
-        }
-      }
-      dispatch(switchContentLanguage)
-      createNavigateToCategory('Dashboard', dispatch, ownProps.navigation)({
```
maxammann: This prop is not passed to the now undefined because only `changeLanguage` is passed to `LanguageNotAvailableContainer` in the hoc.

If I'm right then you can try to fix the flow types here.

src/modules/error/hocs/withError.js
```diff
@@ -1,27 +1,38 @@
...
 
       return <Component {...props} />
     }
   }
-
-  return ErrorComponent
 }
 
-export default withError
+export default withLanguageNotAvailable
```
maxammann: This file should be renamed to `withLanguageNotAvailable`

I think a name like `handleUnavailableLanguage` would be better. Or `withUnavailableLangaugeHandler`.
Or `withUnavailablityHandler` to make because it may not only handle language in the future.

src/locales.json
```diff
@@ -247,18 +249,21 @@
...
       },
       "loadingFailed": "خطایی هنگام بارگیری محتوا رخ داده است.",
-      "pageTitle": "خطا"
+      "pageTitle": "خطا",
+      "generalError": "خطایی رخ داده است"
     },
     "fr": {
-      "chooseYourLanguage": "Choisis ta Langue:"
+      "chooseYourLanguage": "Choisis ta Langue:",
+      "generalError": "Une erreur s'est produite"
```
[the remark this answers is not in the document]
klinzo: Those are already part of the translation sheet, but with the key `ErrorGeneral`

src/modules/error/hocs/withError.js
klinzo: Imo the HOC should be renamed to `withError` instead of renaming the file. Language not available is also an error so `withError` is better fitting.

src/modules/common/containers/LanguageNotAvailableContainer.js
klinzo: `createNavigateToCategory` isn't a prop here, but part of the code that is executed on `changeLanguage`, so it should be fine.

src/modules/error/hocs/withError.js
klinzo: What do you think @Schedulaar

Schedulaar: Well, yeah, I guess I'd go for `withError`

src/routes/events/containers/EventsContainer.js
```diff
@@ -10,29 +10,28 @@ import createNavigateToEvent from '../../../modules/app/createNavigateToEvent'
...
 
-  if (languages && !languages.map(languageModel => languageModel.code).includes(language)) {
-    return {invalidLanguage: true}
-  }
+const mapStateToProps = (state: StateType, ownProps: OwnPropsType) => {
+  const {resourceCache, eventsRouteMapping, city} = state.cityContent
 
-  if (eventsRouteMapping.errorMessage !== undefined) {
+  if (eventsRouteMapping.errorMessage !== undefined || resourceCache.errorMessage !== undefined) {
     return {error: true}
```
Schedulaar: So, do we ignore the `errorMessage` for now? We might want to display it, don't we?

klinzo: There is a separate task to display error codes (https://issues.integreat-app.de/browse/NATIVE-193). I think displaying "cryptic" error message (which these error messages are) does not increase the usability.

PR #71, opened 2019-06-05
NATIVE-88: Add custom fonts 

Schedulaar:


src/modules/theme/constants/fonts.js
```diff
@@ -2,22 +2,43 @@
...
+  contentFontBold: string,
+  webviewFontFamilies: string,
   contentFontSize: string,
   contentLineHeight: number,
   standardParagraphMargin: string,
   subTitleFontSize: string
 }
 
-export default {
+export const arabicFonts: FontsType = { // used for 'ar', 'fa' and 'ku'
```
maxammann: I'm thinking about discussing a font switch on the next conference. I think nobody would notice a change actually...
Google Noto is quite cool when it comes to glyph support: https://www.google.com/get/noto/

maxammann: What do you think? Should I put it on the agenda?

src/modules/theme/containers/CustomThemeProvider.js
```diff
@@ -11,9 +11,11 @@ type PropsType = {
   darkMode: boolean
 }
 
+// todo: remove this provider once styled-components is upgraded to v4.2.2
+// or once https://github.com/styled-components/styled-components/issues/2578 is closed
```
maxammann: Can you create a task for this?

src/modules/theme/hocs/withTheme.js
```diff
@@ -0,0 +1,32 @@
...
+
+const withTheme = <Props> (mapPropsToLanguage?: MapPropsToLanguageType<Props>) =>
+  (Component: React.ComponentType<Props>): React.ComponentType<$Diff<Props, {| theme: ThemeType |}>> => {
+    return class extends React.Component<{}> {
+      static displayName = wrapDisplayName(Component, 'withTheme')
+
+      render () {
+        const language = mapPropsToLanguage && mapPropsToLanguage(this.props)
+        const theme: ThemeType = {
+          colors: brightColors,
```
maxammann: Components which use this HOC will not be 100% testable because the theme is not mockable.

The "right" way to do this would be to add the "theme" to the context. This HOC would get the theme from the context then and not by importing it.

src/modules/theme/hocs/withTheme.js
```diff
@@ -0,0 +1,32 @@
...
+import dimensions from '../constants/dimensions'
+import { arabicFonts, defaultFonts } from '../constants/fonts'
+
+type MapPropsToLanguageType<Props> = Props => string
+
+const ARABIC_LANGUAGES = ['ar', 'fa', 'ku']
+
+const withTheme = <Props> (mapPropsToLanguage?: MapPropsToLanguageType<Props>) =>
+  (Component: React.ComponentType<Props>): React.ComponentType<$Diff<Props, {| theme: ThemeType |}>> => {
+    return class extends React.Component<{}> {
```
maxammann: I'm not sure whether the flow type is correct here. Below you are accessing `this.props` which has the type `Props`. But in this line you are saying the prop type is `{}`. I think `Props` should be the right one. At least `{}` seems weird.

If you didn't have a look yet there is a quite good description: https://flow.org/en/docs/react/hoc/#toc-injecting-props

src/routes/dashboard/containers/DashboardContainer.js
```diff
@@ -61,5 +61,5 @@ export default compose([
   branch(props => props.error, renderComponent(Failure)),
   branch(props => props.invalidLanguage, renderComponent(LanguageNotAvailableContainer)),
   translate('dashboard'),
-  withTheme
+  withTheme(props => props.language)
```
maxammann: It's good that we get rid of `withTheme`. Flow types are bad for it... Like for translate actually.

src/routes/settings/container/SettingsContainer.js
```diff
@@ -27,6 +27,7 @@ const mapStateToProps = (state: StateType) => {
   }
 }
 
+const ThemedSettings = withTheme(props => props.language)(translate('settings')(Settings))
```
maxammann: I think ThemedSettings is not used because it is not exported as default.

src/modules/common/renderHtml.js
```diff
@@ -54,26 +55,67 @@ const renderJS = (files: FileCacheStateType) => `
...
         margin: 0;
         padding: 0;
         
-        /*font-family: \${theme.fonts.contentFontFamily};*/   
+        font-family: ${theme.fonts.webviewFontFamilies};
         font-size: ${theme.fonts.contentFontSize};
         line-height: ${theme.fonts.contentLineHeight};
+        font-size-adjust: ${theme.fonts.fontSizeAdjust};
+        background-color: ${theme.colors.backgroundColor};
         /*\${props => props.centered && css\`
```
maxammann: Can you remember why this is here? If not remove it.

maxammann:
Oh sorry did not see the WIP in the title.

src/modules/theme/constants/fonts.js
Schedulaar: Should this also replace Raleway? I like Raleway though :D

src/modules/theme/containers/CustomThemeProvider.js
Schedulaar: Done

src/modules/common/renderHtml.js
Schedulaar: This is for centering the (former) RemoteContent. I don't remove it now, since we might get back to this.

ios/Integreat.xcworkspace/contents.xcworkspacedata
```diff
@@ -1,6 +1,21 @@
...
+   </FileRef>
+   <FileRef
+      location = "group:../android/app/src/main/assets/fonts/OpenSans-Regular.ttf">
+   </FileRef>
+   <FileRef
+      location = "group:../android/app/src/main/assets/fonts/Raleway-Bold.ttf">
+   </FileRef>
+   <FileRef
+      location = "group:../android/app/src/main/assets/fonts/Raleway-Regular.ttf">
+   </FileRef>
```
Schedulaar: Here's something wrong I guess. Seems like I've imported these files not the right way. Can you check this on the mac @maxammann?

src/modules/app/components/App.js
```diff
@@ -51,16 +49,14 @@ class App extends React.Component<PropsType, AppStateType> {
...
-            <Dialog>
-              <>
-                <AndroidStatusBarContainer />
-                <IOSSafeAreaView>
-                  <Navigator />
-                </IOSSafeAreaView>
-              </>
-            </Dialog>
-          </CustomThemeProvider>
+          <LaunchInquiry>
```
klinzo: :+1:

src/modules/common/components/Page.js
```diff
@@ -112,7 +112,7 @@ class Page extends React.Component<PropType, StateType> {
     const {title, children, content, files, theme, language, cityCode, lastUpdate} = this.props
     const height = this.state.webViewHeight
     const width = this.state.webViewWidth
-
+    const dataDetectorTypes: any = 'all' // flow doesn't recognize this as type of DataDetectorTypes
```
maxammann: Use a FlowFixMe instead of a workaround with any.

ios/Integreat.xcworkspace/contents.xcworkspacedata
maxammann: Not right now, but I think you can revert all the changes in the workspacedata

src/modules/common/components/Page.js
Schedulaar: Flow is super dumb in this case as well, btw. We should think about upgrading it soonish... I hope they improved their stuff a little bit at least :D

PR #72, opened 2019-06-08
NATIVE-254: Scan urls only if necessary

Schedulaar:
A major performance improvement...

### DIFF
--- a/src/modules/endpoint/sagas/loadCityContent.js
+++ b/src/modules/endpoint/sagas/loadCityContent.js
@@ -9,6 +9,8 @@ import fetchResourceCache from './fetchResourceCache'
 import moment from 'moment-timezone'
 import type { PushLanguagesActionType, SetCityContentLocalizationType } from '../../app/StoreActionType'
 import loadLanguages from './loadLanguages'
+import ResourceURLFinder from '../ResourceURLFinder'
+import buildResourceFilePath from '../buildResourceFilePath'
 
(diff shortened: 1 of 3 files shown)


src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -54,16 +56,24 @@ export default function * loadCityContent (
   yield put(pushLanguages)
 
   if (languages.map(language => language.code).includes(newLanguage)) {
-    const [categoryUrls, eventUrls] = yield all([
-      call(loadCategories, newCity, newLanguage, dataContainer, shouldUpdate),
-      call(loadEvents, newCity, newLanguage, dataContainer, shouldUpdate)
+    const [categoriesMap, events] = yield all([
+      call(loadCategories, newCity, newLanguage, dataContainer, shouldUpdate, shouldRefreshResources),
+      call(loadEvents, newCity, newLanguage, dataContainer, shouldUpdate, shouldRefreshResources)
```
klinzo: As far as I can see, the `shouldRefreshResources` is never used in `loadCategories´ and `loadEvents`, so it should probably be removed from the call.

Schedulaar: Whoops, yeah. That's a left-over. Thanks :)

Schedulaar:
@maxammann Forgot to add a description to issue, true...

PR #73, opened 2019-06-08
NATIVE-255: Remove ExtrasStack to give back-button the expected behavior

Schedulaar:


Schedulaar:
flop...

maxammann:
The reason of the flop errors are the HOCs in the containers. Probably a misuse of connect or if translate /withTheme is used.

maxammann:
I added the wip tag because the flop errors are pretty nasty in this one :P

Schedulaar:
Actually, I'm pretty sure, this is a mistake of flow: It says:  
```- Either property \`navigationOptions\` is missing in \`React.ComponentType\` [1] but exists in

   \`withOptionalNavigationOptions\` [2].
```
but navigationOptions is (as the name suggests) an *optional* property....

maxammann:
The flop error message is pretty weird but should not need the wrapper component.

The error displayed in the wrong place and the error message is misleading.
So the error is in one of the following components: ExtrasContainer, WohnenExtraContainer, SprungbrettExtraContainer, ExternalExtraContainer

Probably hocs are used in a wrong way in these components. Like an untyped connect(), translate etc.

Schedulaar:
Feel free to dig into this, but that's not part of this task & stuff works right now...
I'd leave it like this, otherwise flow is using up 80% of productivity...

Schedulaar:
@maxammann I could fix it now -.-'

PR #76, opened 2019-06-13
NATIVE-170: Mapping for city model

Simple-codinger:


src/modules/endpoint/DatabaseConnector.js
```diff
@@ -187,6 +201,40 @@ class DatabaseConnector {
     await this.writeFile(path, JSON.stringify(languages))
   }
 
+  async storeCities (cities: Array<CityModel>) {
+    const jsonModels = cities.map((city: CityModel): ContentCityJsonType => ({
```
klinzo: Couldn't you just call `JSON.stringify()` on the `Array<CityModel>`? I think the `map` here is not necessary.

src/modules/endpoint/DatabaseConnector.js
```diff
@@ -73,11 +83,15 @@ const mapToObject = (map: Map<string, string>) => {
 }
 
 class DatabaseConnector {
-  getContentPath (key: string, context: DatabaseContext): string {
+  getContentPath (key: string, context: DatabaseContext = null): string {
```
klinzo: Should probably be a `context: ?DatabaseContext = null`.

src/modules/endpoint/DatabaseConnector.js
```diff
@@ -73,11 +83,15 @@ const mapToObject = (map: Map<string, string>) => {
...
     if (!key) {
       throw Error('Key mustn\'t be empty')
     }
 
+    if (context == null) {
+      return `${CONTENT_DIR_PATH}/${key}.json`
```
klinzo: Maybe this should rather be excluded to a separate method, since it does not have to do anything with the `context`. Not sure though.

src/modules/endpoint/DatabaseConnector.js
maxammann: A more clear syntax would be: `(key: string, context?: DatabaseContext)`
Then you can just omit the context

src/modules/endpoint/DatabaseConnector.js
maxammann: I think it is fine as the json lies withing the `CONTENT_DIR_PATH`

src/modules/endpoint/DatabaseConnector.js
maxammann: We should explicitly use the `map` here imo. We do not expect the JSON files to change if we change the `CityModel`

src/modules/endpoint/DatabaseConnector.js
```diff
@@ -187,6 +201,40 @@ class DatabaseConnector {
...
+    const jsonModels = cities.map((city: CityModel): ContentCityJsonType => ({
+      name: city.name,
+      live: city.live,
+      code: city.code,
+      extrasEnabled: city.extrasEnabled,
+      eventsEnabled: city.eventsEnabled,
+      sortingName: city.sortingName
+    }))
+
+    await this.writeFile(this.getContentPath('cities'), JSON.stringify((jsonModels)))
```
Schedulaar:
```suggestion
    await this.writeFile(this.getContentPath('cities'), JSON.stringify(jsonModels))
```
one pair of brackets unnecessary :)

src/modules/endpoint/DatabaseConnector.js
maxammann: For consistency with the functions around it you should create a separate funciton

src/modules/endpoint/DatabaseConnector.js
```diff
@@ -187,6 +197,40 @@ class DatabaseConnector {
     await this.writeFile(path, JSON.stringify(languages))
   }
 
+  async storeCities (cities: Array<CityModel>) {
+    const jsonModels = cities.map((city: CityModel): ContentCityJsonType => ({
```
maxammann: The last_update field is missing here. See `storeLastUpdate`.

maxammann:
Alright, seems like we have some problems with this task. The `las_updated` and `languages` are currently not contained in the cities.json
But I'd say we move this functionality to an other task.

For this task it is only important that it does not break the current functionality. This means the `last_updated`  field musn't be overwritten.

src/modules/endpoint/DatabaseConnector.js
```diff
@@ -187,6 +197,40 @@ class DatabaseConnector {
...
+  async loadCities (): Promise<Array<CityModel> | null> {
+    const path = this.getMetaCitiesPath()
+    const fileExists: boolean = await RNFetchblob.fs.exists(path)
+
+    if (!fileExists) {
+      return null
+    }
+
+    const cities = JSON.parse(await this.readFile(path))
+    return cities.map(city =>
```
NKlug: There might also be the case that the `lastUpdate` was already stored for the city, but the rest of the city meta data is missing. So make sure that nothing is `undefined` here.

maxammann:
Closing this for now as it it not clear whether it makes sense to merge develop back in or start over.

PR #77, opened 2019-06-16
NATIVE-110: Link between cities

maxammann:


maxammann:
rdy for review

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -24,71 +25,74 @@ class DefaultDataContainer implements DataContainer {
...
-  getEvents = async (): Promise<Array<EventModel>> => {
-    if (this._context === null) {
-      throw Error('Context has not been set yet.')
-    }
+  getEvents = async (context: DatabaseContext): Promise<Array<EventModel>> => {
+    this.removeOutdated(context)
+
     if (this._events === null) {
-      throw Error('Events are null.')
+      this._events = await this._databaseConnector.loadCategories(context)
```
Simple-codinger: I guess loadEvents() should be called here

maxammann: good catch! you are right

src/modules/endpoint/sagas/loadCategories.js
```diff
@@ -21,19 +22,21 @@ function * loadCategories (
   dataContainer: DataContainer,
   shouldUpdate: boolean
 ): Saga<FetchMapType> {
-  if (!dataContainer.categoriesAvailable() || shouldUpdate) {
+  const context = new DatabaseContext(city, language)
+
+  if (!dataContainer.categoriesAvailable(context) || shouldUpdate) {
```
Schedulaar: The problem now is -- since you completely changed the behavior of the DataContainer --, that `dataContainer.categoriesAvailable` (as well as all other `xyzAvailable` methods) will evaluate to false even if categories are available but not yet loaded from the database.
You might want to revert the DataContainer to that extent, that on every context switch all available data regarding that context is loaded.

src/modules/endpoint/sagas/loadLanguages.js
```diff
@@ -8,12 +8,12 @@ import { baseUrl } from '../constants'
...
-  if (dataContainer.languagesAvailable() && !shouldUpdate) {
+  if (dataContainer.languagesAvailable(city) && !shouldUpdate) {
     console.debug('Using cached languages')
     return
   }
   console.debug('Fetching languages')
   const params = {city}
   const payload: Payload<Array<LanguageModel>> = yield call(() => request(createLanguagesEndpoint(baseUrl), params))
-  yield call(dataContainer.setLanguages, payload.data)
+  yield call(dataContainer.setLanguages, city, payload.data) // fixme
```
Schedulaar: what to fix here?

src/modules/endpoint/sagas/watchFetchEvent.js
```diff
@@ -5,31 +5,40 @@ import { all, call, put, takeLatest } from 'redux-saga/effects'
...
+          path,
+          languages,
+          key,
+          city,
+          language,
+          peek: loadCriterion.peek()
+        }
       }
+      yield put(insert)
     }
```
Schedulaar: `else` do nothing? How does the app react here?

src/modules/layout/components/Header.js
```diff
@@ -112,6 +111,21 @@ class Header extends React.PureComponent<PropsType> {
...
 
+  showLanguageChange (): boolean {
+    const {routeKey, routeMapping} = this.props
+    const route = routeMapping[routeKey]
+
+    if (!route) {
+      return false
+    }
+
+    return !route.peek
```
Schedulaar: I'm still not quite sure why we should disable this on peek

src/modules/layout/components/Header.js
```diff
@@ -112,6 +111,21 @@ class Header extends React.PureComponent<PropsType> {
...
+  }
+
+  showSearch (): boolean {
+    return this.showLanguageChange()
```
Schedulaar: maybe extract a method `isPeekScreen` from `showLanguageChange` and call that for better readability

src/modules/common/components/Thumbnail.js
```diff
@@ -0,0 +1,34 @@
...
+import * as React from 'react'
+
+import { View } from 'react-native'
+import styled from 'styled-components/native'
+import FastImage from 'react-native-fast-image'
+import getFastImageSource from '../getFastImageSource'
+import type { ViewStyleProp } from 'react-native/Libraries/StyleSheet/StyleSheet'
+
+type PropsType = {
+  uri: string | number | null,
```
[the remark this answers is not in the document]
NKlug: Yep. If you 
```import MyPNG from 'myassets/mycoolpng.png```
`MyPNG` will be a number (for some reason).

File: src/modules/endpoint/DefaultDataContainer.js [hunk 8]
```diff
@@ -167,35 +143,42 @@ class DefaultDataContainer implements DataContainer {
...
 
-  categoriesAvailable (): boolean {
+  categoriesAvailable (context: DatabaseContext): boolean {
+    this.removeOutdated(context)
+
     return this._categoriesMap !== null
   }
 
-  languagesAvailable (): boolean {
+  languagesAvailable (city: string): boolean {
```
NKlug: Why do we need the city here? I think you mean something like
```
return this.getLanguagesForCity(city) !== null
```
(with a proper function `getLanguagesForCity` of course)

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -24,71 +25,74 @@ class DefaultDataContainer implements DataContainer {
     this._databaseConnector = new DatabaseConnector()
   }
 
+  removeOutdated (newContext: DatabaseContext) {
```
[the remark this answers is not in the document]
NKlug: +1 for renaming. Which outdated things are removed?

src/modules/endpoint/DataContainer.js
```diff
@@ -30,83 +24,83 @@ export interface DataContainer {
...
    */
-  getLanguages: () => Promise<Array<LanguageModel>>,
+  getLanguages: (context: DatabaseContext) => Promise<Array<LanguageModel>>,
 
   /**
    * Sets the languages and persists them.
    * @param languages
    */
-  setLanguages: (languages: Array<LanguageModel>) => Promise<void>,
+  setLanguages: (city: string, languages: Array<LanguageModel>) => Promise<void>,
```
NKlug: Please adjust the documentation of the functions you touched in this file.

src/modules/endpoint/sagas/loadCategories.js
NKlug: Also, why is this change in architecture necessary? Is there no other way to work with the current DataContainer?

maxammann:
Note: I need to make sure the fast image Thumbnail is used at all occurences

src/modules/layout/components/Header.js
maxammann: Because we do not want to allow changing the language when peeking a screen. Even though it would be technically possible it is not defined what should happen.

It is also not sure which instance to search in.

maxammann: Behavior could be changed in a separate task if we decide to allow this functionality.

src/modules/endpoint/sagas/watchFetchEvent.js
maxammann: If `allContentLoaded` is false we are not able to put events, categories etc. in the state.

This usually happens because the language was not available. In that case it is needed to trigger a refetch

src/modules/endpoint/DataContainer.js
```diff
@@ -30,83 +24,83 @@ export interface DataContainer {
    * Returns an Array of LanguageModels.
    * @throws Will throw an error if the array is null.
    */
-  getLanguages: () => Promise<Array<LanguageModel>>,
+  getLanguages: (context: DatabaseContext) => Promise<Array<LanguageModel>>,
```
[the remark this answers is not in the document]
maxammann: Context makes it compatible with the new caching mechanism.

src/modules/endpoint/DefaultDataContainer.js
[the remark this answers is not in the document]
maxammann: Function is removed now

src/modules/endpoint/DefaultDataContainer.js
maxammann: Languages depend on the city. This is needed now.

maxammann:
@NKlug the change is needed so we can load two cities at the same time like with a real database.

src/modules/endpoint/sagas/loadCategories.js
maxammann: Should be fine now :)

src/modules/endpoint/Cache.js
```diff
@@ -0,0 +1,68 @@
...
+    const value = this.value
+    if (!value) {
+      const newValue: T | null = await this.load(this.databaseConnector, context)
+
+      if (!newValue) {
+        throw new Error('Data is not available in cache and on disk! Therefore getting data from the cache failed.')
+      }
+
+      this.value = newValue
+      return newValue
```
Schedulaar: You might want to set the context here, right?

src/modules/endpoint/ContentLoadCriterion.js
```diff
@@ -0,0 +1,38 @@
+// @flow
+
+import moment from 'moment-timezone'
+import Moment from 'moment'
+
+const MAX_CONTENT_AGE = 24
+
+export type ContentLoadCriterionType = {|
+  peek?: boolean,
```
Schedulaar: Is peek really a content load criterion and not a property which some screen should open with? :thinking:

src/modules/endpoint/ContentLoadCriterion.js
```diff
@@ -0,0 +1,38 @@
...
+    return this._forceUpdate || !lastUpdate ||
+      lastUpdate.isBefore(moment.tz('UTC').subtract(MAX_CONTENT_AGE, 'hours'))
+  }
+
+  peek (): boolean {
+    return this._peek
+  }
+
+  shouldRefreshResources (): boolean {
+    return this._shouldRefreshResources || this._peek
```
Schedulaar: Why should we refresh resources if we're peeking? I'd say should explicitly *not* do it, because we won't need most of it?

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -8,20 +8,48 @@ import type { DataContainer } from './DataContainer'
...
-  _categoriesMap: CategoriesMapModel | null
-  _languages: Array<LanguageModel> | null
-  _resourceCache: CityResourceCacheStateType | null
-  _events: Array<EventModel> | null
 
   constructor () {
     this._databaseConnector = new DatabaseConnector()
+
+    this.caches = {
+      'events': new Cache<Array<EventModel>>(this._databaseConnector,
```
Schedulaar: Why don't you use:
```suggestion
      events: new Cache<Array<EventModel>>(this._databaseConnector,
```
? So why do you need the single-quotes?

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -31,133 +59,84 @@ class DefaultDataContainer implements DataContainer {
...
 
-  getLanguages = async (): Promise<Array<LanguageModel>> => {
-    if (this._context === null) {
-      throw Error('Context has not been set yet.')
-    }
-    if (this._languages === null) {
-      throw Error('Languages are null.')
-    }
-    return this._languages
+  getLanguages = async (context: DatabaseContext): Promise<Array<LanguageModel>> => {
```
Schedulaar: From here on you use `async`, in the above occurences you don't.

src/modules/endpoint/sagas/loadCategories.js
Schedulaar: This is still a problem

src/modules/endpoint/DataContainer.js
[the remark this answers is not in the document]
klinzo: Does this mean if you try to get the language with a different context, i.e. a different language, this will return null even though it should be the same? E.g. if you set the languages with context `augsburg, de` and load the languages with context `augsburg en`?.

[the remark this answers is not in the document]
klinzo: Same for `getCities`

src/modules/endpoint/reducers/morphContentLanguage.js
```diff
@@ -40,11 +40,13 @@ const categoryRouteTranslator = (newCategoriesMap: CategoriesMapModel, newLangua
...
       allAvailableLanguages,
-      language: newLanguage
+      language: newLanguage,
+      city,
+      peek
     }
   }
 
-const eventRouteTranslator = (newEvents: Array<EventModel>, newLanguage: string) =>
+const eventRouteTranslator = (newEvents: Array<EventModel>, city: string, newLanguage: string) =>
```
klinzo: Right now city isn't used

src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -9,32 +9,45 @@ import type {
...
-      params: {
-        categoriesMap,
-        resourceCache,
-        path,
-        languages,
-        depth,
-        key,
-        city,
-        language
+      let languages = [new LanguageModel(language, language)]
```
klinzo: This looks quite wrong.

src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -9,32 +9,45 @@ import type {
...
+
+      if (!loadCriterion.peek()) {
+        languages = yield call(dataContainer.getLanguages, context)
```
klinzo: Why only load the languages if we aren't peeking?

src/modules/endpoint/sagas/watchFetchEvent.js
```diff
@@ -5,31 +5,38 @@ import { all, call, put, takeLatest } from 'redux-saga/effects'
...
+import { ContentLoadCriterion } from '../ContentLoadCriterion'
+import DatabaseContext from '../DatabaseContext'
 
 function * fetchEvent (dataContainer: DataContainer, action: FetchEventActionType): Saga<void> {
-  const {city, language, path, key, forceUpdate, shouldRefreshResources} = action.params
+  const {city, language, path, key, criterion} = action.params
   try {
-    yield call(loadCityContent, dataContainer, city, language, forceUpdate, shouldRefreshResources)
+    const loadCriterion = new ContentLoadCriterion(criterion)
+    const languageLoaded = yield call(loadCityContent,
```
klinzo: In fact this is rather a `cityContentLoaded` and not a `languageLoaded`

src/modules/endpoint/sagas/watchFetchEvent.js
klinzo: So no loading for an infinite time anymore? I had that problem when removing the `throw Error`

src/modules/endpoint/DefaultDataContainer.js
maxammann: I removed async where I didn't need it.

src/modules/endpoint/reducers/morphContentLanguage.js
maxammann: Removed, weird that there was no warning

src/modules/endpoint/sagas/watchFetchCategory.js
maxammann: If we are peeking there is no need for the languages as the language switch is disabled.

The languages of the opened city will be in the state and not those of the peeked city.

src/modules/endpoint/sagas/watchFetchEvent.js
maxammann: Right now loadCityContent only returns false if the `language` is not available in the city. Therefore the name `languageLoaded` is oke.

maxammann: Oke I changed it :P

src/modules/endpoint/sagas/watchFetchCategory.js
maxammann: I fixed this by cleaning a bit up! Thanks for the note!

src/modules/endpoint/DataContainer.js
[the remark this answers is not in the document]
maxammann: Yes, you are right, the languages depend now not only on the city, but also on the language

src/modules/endpoint/ContentLoadCriterion.js
maxammann: I renamed the functions a bit

src/modules/endpoint/sagas/loadCategories.js
maxammann: Should be solved finally :)

src/modules/endpoint/sagas/watchFetchEvent.js
klinzo: :grimacing:

src/modules/endpoint/DataContainer.js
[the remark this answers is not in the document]
klinzo: But that doesn't make sense... that way we have to load it separately every time, don't we?

src/modules/endpoint/DatabaseConnector.js
```diff
@@ -271,6 +269,26 @@ class DatabaseConnector {
     await this.writeFile(path, JSON.stringify(json))
   }
 
+  isCategoriesPersisted (context: DatabaseContext): Promise<boolean> {
```
klinzo: Grammatically this should be `areCategoriesPersisted` :stuck_out_tongue:

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -31,133 +64,86 @@ class DefaultDataContainer implements DataContainer {
...
-  getCategoriesMap = async (): Promise<CategoriesMapModel> => {
-    if (this._context === null) {
-      throw Error('Context has not been set yet.')
-    }
-    if (this._categoriesMap === null) {
-      throw Error('CategoriesMap is null.')
-    }
-    return this._categoriesMap
+  getCategoriesMap = (context: DatabaseContext): Promise<CategoriesMapModel> => {
+    const cache: Cache<Array<EventModel>> = this.caches['categories']
```
klinzo: Why don't you just use `this.caches.categories`? That way flow should be able to check the types.

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -31,133 +64,86 @@ class DefaultDataContainer implements DataContainer {
...
-    if (this._resourceCache === null) {
-      throw Error('CityResourceCache is null.')
-    }
-    if (!this._resourceCache[this._context.languageCode]) {
+  getResourceCache = async (context: DatabaseContext): Promise<LanguageResourceCacheStateType> => {
+    const cache: Cache<CityResourceCacheStateType> = this.caches['resourceCache']
+    const resourceCache = await cache.get(context)
+
+    if (!resourceCache[context.languageCode]) {
+      // TODO: Should be cought properly
```
klinzo:
```suggestion
      // TODO: Should be caught properly
```

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -167,36 +153,29 @@ class DefaultDataContainer implements DataContainer {
...
-  resourceCacheAvailable (): boolean {
-    return this._resourceCache !== null
+  async eventsAvailable (context: DatabaseContext): Promise<boolean> {
+    return this.isCached('events', context) || this._databaseConnector.isEventsPersisted(context)
   }
 
-  lastUpdateAvailable (): boolean {
-    return this._lastUpdate !== null
+  async resourceCacheAvailable (context: DatabaseContext): Promise<boolean> {
+    // FIXME: this function needs to check whether the language exists in the resource cache
```
klinzo: Reminder

src/modules/endpoint/sagas/loadCategories.js
```diff
@@ -21,19 +22,22 @@ function * loadCategories (
   dataContainer: DataContainer,
   shouldUpdate: boolean
 ): Saga<FetchMapType> {
-  if (!dataContainer.categoriesAvailable() || shouldUpdate) {
+  const context = new DatabaseContext(city, language)
+  const categoriesAvailable = yield call({context: dataContainer, fn: dataContainer.categoriesAvailable}, context)
```
klinzo:
```suggestion
  const categoriesAvailable = yield call(dataContainer.categoriesAvailable, context)
```

src/modules/endpoint/sagas/loadEvents.js
```diff
@@ -18,7 +19,10 @@ function * fetchEvents (city: string, language: string): Saga<Array<EventModel>>
 function * loadEvents (
   city: string, language: string, dataContainer: DataContainer, shouldUpdate: boolean
 ): Saga<FetchMapType> {
-  if (!dataContainer.eventsAvailable() || shouldUpdate) {
+  const context = new DatabaseContext(city, language)
+  const eventsAvailable = yield call({context: dataContainer, fn: dataContainer.eventsAvailable}, context)
```
klinzo:
```suggestion
  const eventsAvailable = yield call(dataContainer.eventsAvailable, context)
```

src/modules/endpoint/sagas/loadLanguages.js
```diff
@@ -6,14 +6,17 @@ import { createLanguagesEndpoint, LanguageModel, Payload } from '@integreat-app/
 import request from '../request'
 import { baseUrl } from '../constants'
 import type { DataContainer } from '../DataContainer'
+import DatabaseContext from '../DatabaseContext'
 
-export default function * loadLanguages (city: string, dataContainer: DataContainer, shouldUpdate: boolean): Saga<void> {
-  if (dataContainer.languagesAvailable() && !shouldUpdate) {
+export default function * loadLanguages (context: DatabaseContext, dataContainer: DataContainer, shouldUpdate: boolean): Saga<void> {
+  const languagesAvailable = yield call({context: dataContainer, fn: dataContainer.languagesAvailable}, context)
```
klinzo:
```suggestion
  const languagesAvailable = yield call(dataContainer.languagesAvailable, context)
```

src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -9,32 +9,66 @@ import type {
...
+  // If there are no loaded languages the result is an empty map because we do not have a root category
+  return new Map<string, string>()
+}
 
 function * fetchCategory (dataContainer: DataContainer, action: FetchCategoryActionType): Saga<void> {
-  const {city, language, path, depth, key, forceUpdate, shouldRefreshResources} = action.params
+  const { city, language, path, depth, key, criterion } = action.params
   try {
-    yield call(loadCityContent, dataContainer, city, language, forceUpdate, shouldRefreshResources)
+    const loadCriterion = new ContentLoadCriterion(criterion)
```
klinzo: Why are you creating a new `ContentLoadCriterion` here?

src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -9,32 +9,66 @@ import type {
...
-        city,
-        language
+      const context = new DatabaseContext(city, language)
+      const [categoriesMap] = yield all([
+        call(dataContainer.getCategoriesMap, context)
+      ])
+
+      let resourceCache = {}
+
+      const resourceCacheAvailable = yield call({context: dataContainer, fn: dataContainer.resourceCacheAvailable}, context)
```
klinzo:
```suggestion
      const resourceCacheAvailable = yield call(dataContainer.resourceCacheAvailable, context)
```

src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -9,32 +9,66 @@ import type {
...
+
+      if (resourceCacheAvailable) {
+        // TODO: This call should happen paralel
```
klinzo: Shouldn't we create an issue or do it in this task?

src/modules/endpoint/sagas/watchFetchEvent.js
```diff
@@ -5,31 +5,40 @@ import { all, call, put, takeLatest } from 'redux-saga/effects'
...
+      dataContainer, city, language,
+      loadCriterion
+    )
+    if (cityContentLoaded) {
+      // Only proceed if the content is ready to be pushed to the state. If not then the UI automatically displays an
+      // appropriate error
+      const context = new DatabaseContext(city, language)
+      const [events, resourceCache, languages] = yield all([
+        call(dataContainer.getEvents, context),
+        call(dataContainer.getResourceCache, context),
```
klinzo: You should probably check whether the `resourceCache` is available and load the languages the same way as you did in `watchFetchCategory` if it is permitted to `peek` events.

src/modules/endpoint/sagas/loadCategories.js
maxammann: Availability checks work now. I promise! :)

src/modules/endpoint/sagas/loadLanguages.js
maxammann: Would not work that way because languagesAvailable references this.

src/modules/endpoint/sagas/watchFetchCategory.js
maxammann: There is a class and a plain js object version of it.

src/modules/endpoint/sagas/watchFetchCategory.js
maxammann: Fixed this

src/modules/endpoint/sagas/watchFetchEvent.js
maxammann: Events are not part of this PR. I'll create a task for them.

src/modules/endpoint/DefaultDataContainer.js
maxammann: Remove it :)

src/modules/endpoint/DefaultDataContainer.js
maxammann: Done

src/modules/endpoint/DatabaseConnector.js
maxammann: Yeah but I went with is so it is consitent with the other functions.

maxammann:
@klinzo @Schedulaar I'm finally done :) After some careful changes it works easier :)

src/modules/endpoint/DataContainer.js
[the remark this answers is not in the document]
maxammann: Langauges actually depend on the city and on the language. So any change in the context means a fetch is required.

src/modules/endpoint/sagas/loadEvents.js
```diff
@@ -18,7 +19,10 @@ function * fetchEvents (city: string, language: string): Saga<Array<EventModel>>
 function * loadEvents (
   city: string, language: string, dataContainer: DataContainer, shouldUpdate: boolean
 ): Saga<FetchMapType> {
-  if (!dataContainer.eventsAvailable() || shouldUpdate) {
+  const context = new DatabaseContext(city, language)
+  const eventsAvailable = yield call({ context: dataContainer, fn: dataContainer.eventsAvailable }, context)
```
Schedulaar: Still, I think it's easier to just define those as arrow-functions...

src/modules/endpoint/ContentLoadCriterion.js
```diff
@@ -0,0 +1,43 @@
...
+  forceUpdate: boolean,
+  shouldRefreshResources: boolean
+|}
+
+export class ContentLoadCriterion {
+  _forceUpdate: boolean
+  _shouldRefreshResources: boolean
+  _peeking: boolean
+
+  constructor ({ forceUpdate, shouldRefreshResources }: ContentLoadCriterionType, peeking: boolean) {
```
Schedulaar: Why don't you add `peeking: boolean` to ContentLoadCriterionType? :thinking:

src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -1,40 +1,80 @@
...
-        resourceCache,
-        path,
-        languages,
-        depth,
-        key,
-        city,
-        language
+    if (cityContentLoaded) {
+      // Only proceed if the content is ready to be pushed to the state. If not then the UI automatically displays an
+      // appropriate error
```
Schedulaar: For simplicity you could instead check `!cityContentLoaded` and return here...

src/modules/endpoint/sagas/watchFetchEvent.js
```diff
@@ -5,31 +5,40 @@ import { all, call, put, takeLatest } from 'redux-saga/effects'
...
   try {
-    yield call(loadCityContent, dataContainer, city, language, forceUpdate, shouldRefreshResources)
+    const loadCriterion = new ContentLoadCriterion(criterion, false)
+    const cityContentLoaded = yield call(loadCityContent,
+      dataContainer, city, language,
+      loadCriterion
+    )
+    if (cityContentLoaded) {
+      // Only proceed if the content is ready to be pushed to the state. If not then the UI automatically displays an
+      // appropriate error
```
Schedulaar: Same as above

src/modules/layout/components/Header.js
```diff
@@ -61,11 +61,35 @@ type PropsType = {|
   scenes: Array<NavigationScene>,
   t: TFunction,
   theme: ThemeType,
+  peeking: boolean | 'unsure',
```
Schedulaar: No idea what `unsure` means in this context. Does it mean `unknown'?

src/modules/layout/components/Header.js
```diff
@@ -61,11 +61,35 @@ type PropsType = {|
...
+class Header extends React.PureComponent<PropsType, StateType> {
+  constructor () {
+    super()
+
+    this.state = {
+      peeking: true
+    }
+  }
+
+  static getDerivedStateFromProps (props: PropsType, state: StateType): StateType {
```
Schedulaar: It should be pretty easy to get rid of this function. Since it's not necessary and `getDerivedStateFromProps` should be used rarely, we should avoid it here.

src/routes/events/components/EventListItem.js
```diff
@@ -33,7 +33,7 @@ class EventListItem extends React.PureComponent<PropsType> {
 
   render () {
     const { event, language, navigateToEvent, theme } = this.props
-    const thumbnail = event.thumbnail ? { uri: event.thumbnail } : this.getEventPlaceholder(event.path.length)
+    const thumbnail = event.thumbnail ? event.thumbnail : this.getEventPlaceholder(event.path.length)
```
Schedulaar: Can't you use 
```suggestion
    const thumbnail = event.thumbnail || this.getEventPlaceholder(event.path.length)
```

src/modules/layout/components/Header.js
maxammann: Provide me with an other way of doing this :) The input of the props is "cached" in the state. There is only this way to do it

src/modules/layout/components/Header.js
maxammann: It is like: true, false, unsure
or: ja, nein, jain :)

src/modules/endpoint/ContentLoadCriterion.js
maxammann: Becaues ContentLoadCriterionType is used elsewhere without the peeking property.

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -31,109 +64,56 @@ class DefaultDataContainer implements DataContainer {
...
-    }
-    await this._databaseConnector.storeCategories(categories, this._context)
-    this._categoriesMap = categories
+  setCategoriesMap = async (context: DatabaseContext, categories: CategoriesMapModel) => {
+    const cache: Cache<CategoriesMapModel> = this.caches['categories']
+    await cache.cache(categories, context)
   }
 
-  setCities = async (cities: Array<CityModel>) => {
+  setCities = (cities: Array<CityModel>) => {
```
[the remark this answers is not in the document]
maxammann: Because it is unnecessary here. Only if you use the async/await syntax it makes sense afaik.

src/modules/layout/components/Header.js
maxammann: If peeking is unsure then we the Header just has no information whether it is currently peeking. In that case it uses the cached value from the local state.

src/modules/endpoint/DataContainer.js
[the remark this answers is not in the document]
maxammann: Hm weird it loads an url like: https://cms.integreat-app.de/augsburg/de/wp-json/extensions/v3/languages

src/modules/layout/components/Header.js
Schedulaar: Well the exact same behavior is not possible (but actually it is): Once we know that the route is peeking or not this property doesn't change anymore (especially it's never going to be set to `unsure` anyhow). You could just make `peeking` a nullable prop that defaults to `true` (or decide in the container to use `true` if it's unsure). Then you would get the same behavior as far as I can see...

src/modules/endpoint/ContentLoadCriterion.js
klinzo: Shouldn't it just be `false` then?

src/modules/endpoint/DataContainer.js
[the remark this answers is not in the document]
klinzo: Yep, this is an cms issue. We always load the languages with `de` because `de` always exists, that's why you don't have to pass a language to the languages endpoint. So fetching the languages for every language is just unnecessary. Same goes for `getCities`

src/modules/endpoint/DefaultDataContainer.js
klinzo: Not really done, right? :thinking:

klinzo: Same for other getters

src/modules/endpoint/sagas/loadCategories.js
```diff
@@ -21,19 +22,22 @@ function * loadCategories (
   dataContainer: DataContainer,
   shouldUpdate: boolean
 ): Saga<FetchMapType> {
-  if (!dataContainer.categoriesAvailable() || shouldUpdate) {
+  const context = new DatabaseContext(city, language)
+  const categoriesAvailable = yield call({ context: dataContainer, fn: dataContainer.categoriesAvailable }, context)
+
+  if (!categoriesAvailable || shouldUpdate) {
     // data is already loaded and should not be updated
```
klinzo: Not true, is it?

src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -11,80 +11,90 @@ import type { FetchCitiesActionType, InitializeCityContentActionType } from '../
...
+
+    yield put(initializeCityContent)
 
-  const initializeCityContent: InitializeCityContentActionType = {
-    type: 'INITIALIZE_CITY_CONTENT',
-    params: {
-      city: newCity,
-      language: newLanguage,
-      languages
+    if (!languages.map(language => language.code).includes(newLanguage)) {
```
klinzo: We should  maybe keep in mind that it will not be checked whether the linked language is available. This should probably be fixed in NATIVE-249

src/modules/endpoint/sagas/watchFetchEvent.js
klinzo: Reminder

src/modules/endpoint/selectors/isPeekingRoute.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * This fetch corresponds to a peek if the major content city is not equal to the city of the current route.
+ * In this case the fetching and header behaves different. It doesn't fetch resources for example.
+ *
+ * @param state The current state
+ * @param params The params which contain the routeCity
+ * @returns true if the fetch corresponds to a peek
+ */
+const isPeekingRoute: InputSelector<StateType, SelectPropsType, boolean> = (
```
klinzo: Couldn't you just supply peeking when pushing the route?

src/modules/layout/components/Header.js
klinzo: Why don't you use `null` instead  of `unsure`?

src/modules/endpoint/selectors/isPeekingRoute.js
maxammann: That's how it was before this. It actually is easier this way. We had redundant information in the redux store before this.

src/modules/endpoint/sagas/loadCityContent.js
maxammann: Which linked language? `newLanguage` will have the value of the linked language.

src/modules/layout/components/Header.js
maxammann: Maybe I could, yes but I like that way of expressing a trinary more.

src/modules/endpoint/ContentLoadCriterion.js
maxammann: `peeking` can be true or false here.

src/modules/categories/components/CategoryListItem.js
```diff
@@ -66,8 +66,7 @@ class CategoryListItem extends React.Component<PropsType> {
       <Row>
         <StyledLink onPress={this.onCategoryPress} underlayColor={this.props.theme.colors.backgroundAccentColor}>
           <>
-            <CategoryThumbnail source={category.thumbnail ? { uri: category.thumbnail } : iconPlaceholder}
-                               resizeMode={FastImage.resizeMode.contain} />
+            <CategoryThumbnail source={category.thumbnail ? category.thumbnail : iconPlaceholder} />
```
maxammann:
```suggestion
            <CategoryThumbnail source={category.thumbnail || iconPlaceholder} />
```

src/modules/endpoint/sagas/watchFetchEvent.js
maxammann: It's on my Integreat todo ;)

src/modules/endpoint/DataContainer.js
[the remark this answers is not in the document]
maxammann: Hm I don't think this is a cms issue after all. For example if the cms would be configured right:

`augsburg/de/wp-json/extensions/v3/languages` would return "Englisch"
`augsburg/en/wp-json/extensions/v3/languages` would return "English"

src/modules/layout/components/Header.js
maxammann: I'll look over it but not sure whether you are right.

src/modules/endpoint/DataContainer.js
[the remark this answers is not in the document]
Schedulaar: No this is something we definitely decided on a few year(s) ago! We always load 'de' since they might even return inconsistent language sets and the one configured in the `de`-version is always correct! Besides, the "title" of the language is always translated in the Native Language itself...

maxammann:
As far as I can see only our caching the database is left. So getCities should not depend on DatabaseContext. getLanguages() should depend on DatabaseContext imo.

src/modules/endpoint/DefaultDataContainer.js
maxammann: It actually is done, couldn't push because Github was down...

src/modules/endpoint/DataContainer.js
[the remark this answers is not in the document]
maxammann: Hm yeah not sure. Long time, maybe a commit message documents this but I'm not sure ^^

But imo it makes no sense what we decided earlier. I also communicated the Django team half a year ago that the CMS delivers the languages based on the current language.

**For this PR, I would make it depend only on the city because of the api-client.**

src/modules/endpoint/sagas/loadCityContent.js
klinzo: Yep, but we can't be sure that language is available in the linked city

src/modules/endpoint/selectors/isPeekingRoute.js
klinzo: Allright

src/modules/endpoint/DefaultDataContainer.js
klinzo: Oh okay

src/modules/endpoint/sagas/loadCityContent.js
maxammann: Oh oke now I think I know what you mean. For example if you put XX as language then it will probably fail here.

Yeah should test this maybe. How did you do handle this with the linking within a city?

maxammann:
I'll merge now and resolve the remaining issues later on.

src/modules/endpoint/sagas/loadCityContent.js
klinzo: Not at all, there is a task for this. I think https://issues.integreat-app.de/browse/NATIVE-249

PR #81, opened 2019-06-24
NATIVE-53: Fix flow for connect()

klinzo:


src/modules/app/components/Navigator.js
```diff
@@ -74,7 +76,7 @@ export const AppStack = createStackNavigator(
 
 export const LandingStack = createSwitchNavigator(
   {
-    //            The translate() HOC does not calculate props correctly right now. Because the connect
+    // The translate() HOC does not calculate props correctly right now. Because the connect
     // $FlowFixMe in LandingContainer needs the translation HOC there is a flow error here. (NATIVE-53)
```
klinzo: @maxammann any idea what to do about this? Haven't had time yet to fix this (also the two issues above). If not, I'll try to figure it out. But sounds like you had an idea already here.

flow-typed/npm/@mapbox/react-native-mapbox-gl_vx.x.x.js
```diff
@@ -0,0 +1,263 @@
+// flow-typed signature: f658ad0d9bed2e875e2884e6ed9fd396
```
maxammann: I removed all unnecessary stubs some time ago. I'm quite sure these stubs are not required. Therefore I'd say we remove them for now.

src/modules/app/components/Navigator.js
maxammann: I will look into them but these should be fixed in this PR :)

src/modules/app/createNavigateToEvent.js
```diff
@@ -22,7 +22,7 @@ export default (dispatch: Dispatch<StoreActionType>, navigation: NavigationScree
     key
   })
 
-  return dispatch({
+  dispatch({
```
maxammann: I think we want to return the dispatch here. Was there a error with a return type?

src/modules/common/containers/LanguageNotAvailableContainer.js
```diff
@@ -2,10 +2,10 @@
 
 import LanguageNotAvailablePage from '../components/LanguageNotAvailablePage'
 import withTheme from '../../theme/hocs/withTheme'
-import compose from 'lodash/fp/compose'
```
maxammann: What do you think about restricting this import? https://eslint.org/docs/rules/no-restricted-imports

Would create  a task for this :)

src/modules/i18n/containers/I18nProviderContainer.js
```diff
@@ -0,0 +1,23 @@
...
+
+type PropsType = {| ...OwnPropsType, ...DispatchPropsType |}
+
+const mapDispatchToProps = (dispatch: Dispatch<StoreActionType>): DispatchPropsType => ({
+  setUiDirection: (direction: UiDirectionType) => {
+    dispatch(setUiDirection(direction))
+  }
+})
+
+export default connect<PropsType, OwnPropsType, _, _, _, _>(() => ({}), mapDispatchToProps)(I18nProvider)
```
maxammann:
```suggestion
export default connect<PropsType, OwnPropsType, _, _, _, _>(undefined, mapDispatchToProps)(I18nProvider)
```

I think passing `undefined` is better.

src/modules/layout/components/Header.js
```diff
@@ -58,15 +58,11 @@ const MaterialHeaderButtons = props => {
 }
 
 type PropsType = {|
+  navigation: NavigationScreenProp<*>,
```
maxammann: Can you try there to replace:
https://github.com/Integreat/integreat-react-native-app/blob/a0fb69e33ae78d8be23242b57b47a4f4326d1f10/src/modules/layout/components/Header.js#L85

with `this.props.navigation`? Would fix quite a few flow errors in this file. Also it would be a validation if Header really has the navigation prop.

src/routes/search/containers/SearchModalContainer.js
```diff
@@ -15,7 +15,7 @@ type OwnPropsType = {| navigation: NavigationScreenProp<*> |}
 
 export type PropsType = {|
   categories: CategoriesMapModel | null,
-  navigateToCategory: NavigateToCategoryParamsType => StoreActionType,
+  navigateToCategory: NavigateToCategoryParamsType => void,
```
maxammann: I wrote earlier that you should return the result of the dispatch. I think this is the type you changed to match this.

Not sure if this return type makes sense or not. Maybe void is better.

src/modules/app/createNavigateToEvent.js
maxammann: Maybe I'm wrong: https://github.com/Integreat/integreat-react-native-app/pull/81#discussion_r296856858

maxammann:
The fix was quite easy. Probably you forgot to push some flow types which cause these problems

maxammann:
Also make sure that there is only one flow types file for react-i18next. I think there are currently two.

src/modules/common/containers/LanguageNotAvailableContainer.js
klinzo: I think it would not be wrong, but we should also restrict the `compose` of `lodash`.

src/routes/search/containers/SearchModalContainer.js
klinzo: I think it's not really important if we return the action or not, as long as its's consistent. Imo it looks a bit cleaner without the returned action, since it's just a general `StoreActionType` anyway. If we want return types, it should be the actual `ActionType`.

klinzo:
@maxammann Thanks!

src/modules/common/containers/LanguageNotAvailableContainer.js
klinzo: I'll create a task.

src/modules/layout/components/Header.js
klinzo: Done!

src/routes/search/containers/SearchModalContainer.js
klinzo: Alright, @maxammann and I think both it's good as it is.

src/routes/dashboard/components/Dashboard.js
```diff
@@ -18,29 +18,33 @@ import type { NavigateToCategoryParamsType } from '../../../modules/app/createNa
...
   t: TFunction
-}
+|}
 
 class Dashboard extends React.Component<PropsType> {
-  getNavigationTileModels (): Array<TileModel> {
-    const {t, cityCode, language, navigateToCategory} = this.props
+  getNavigationTileModels (cityCode: string, language: string): Array<TileModel> {
+    const {navigation, navigateToCategory, navigateToEvent, t} = this.props
+    const extras = () => {
```
Schedulaar:
```suggestion
    const navigateToExtras = () => {
```

src/routes/dashboard/components/Dashboard.js
```diff
@@ -55,39 +59,29 @@ class Dashboard extends React.Component<PropsType> {
         path: 'extras',
         thumbnail: offersIcon,
         isExternalUrl: false,
-        onTilePress: this.extras,
+        onTilePress: extras,
```
Schedulaar:
```suggestion
        onTilePress: navigateToExtras,
```

PR #86, opened 2019-06-28
NATIVE-263: Persist content language

klinzo:


src/modules/app/StateType.js
```diff
@@ -68,38 +68,27 @@ export const defaultCitiesState: CitiesStateType = {
 }
 
 export type SearchRouteType = {|
-  +categoriesMap: CategoriesMapModel | null
+  +categoriesMap: CategoriesMapModel
 |}
 
 export type CityContentStateType = {|
-  +lastUpdate: Moment | null,
```
NKlug: Where/how do we deal with the lastUpdate now?

klinzo: Just like it was before since this is just the redux state. We didn't even put any values there or read from it, so I figured it can be removed.

src/modules/app/createReduxStore.js
```diff
@@ -51,6 +53,7 @@ const createReduxStore = (
     darkMode: false,
 
     cities: defaultCitiesState,
+    contentLanguage: DEFAULT_LANGUAGE,
```
maxammann: Probably would be cleaner if a defaultContentLanguage should be introduced in the StateType like defaultCitietsState and set it to DEFAULT_LANGUAGE.

src/modules/endpoint/reducers/__tests__/morphContentLanguage.spec.js
```diff
@@ -291,15 +283,16 @@ describe('morphContentLanguage', () => {
   }): CityContentStateType => {
     let state = initialState
 
-    const setCityContentLocalizationAction: SetCityContentLocalizationType = {
-      type: 'SET_CITY_CONTENT_LOCALIZATION',
+    const initializeCityContentAction: InitializeCityContentActionType = {
+      type: 'INITIALIZE_CITY_CONTENT',
```
maxammann: Rename to SET so we stay more consistent with the naming of the other actions.

src/modules/endpoint/reducers/initializeCityContent.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+
+import type { CityContentStateType } from '../../app/StateType'
+import type { InitializeCityContentActionType } from '../../app/StoreActionType'
+
+const initializeCityContent = (
```
maxammann: The naming of this can stay initializeCityContent as it actually initialized it from an null/undefined state.

src/modules/i18n/components/I18nProvider.js
```diff
@@ -75,11 +79,17 @@ class I18nProvider extends React.Component<PropsType, StateType> {
...
+    let language: ?string = await this.languageAsyncStorage.loadLanguage()
+    if (!language) {
+      language = this.getI18nextLanguage()
+      await this.languageAsyncStorage.setLanguage(language)
+    }
 
-    this.setState({language: targetLanguage})
-    this.props.setUiDirection(RTL_LANGUAGES.includes(targetLanguage) ? 'rtl' : 'ltr')
+    this.setState({ language })
+    setUiDirection(RTL_LANGUAGES.includes(language) ? 'rtl' : 'ltr')
```
maxammann: These two lines affect the UI language. The content in the languageAsyncStorage (the content language) should not affect the ui language.

src/modules/layout/components/Header.js
```diff
@@ -134,6 +129,10 @@ class Header extends React.PureComponent<PropsType> {
     const { navigation, t, theme } = this.props
     const sharePath = navigation.getParam('sharePath')
 
+    if (!this.props.navigation) {
+      return null
```
maxammann: Is this expected to happen? If not probably a Error should be thrown.

src/modules/endpoint/reducers/__tests__/morphContentLanguage.spec.js
klinzo: Like `SET_CITY_CONTENT`? Imo it is not clear what it is then.

src/modules/common/components/LanguageNotAvailablePage.js
```diff
@@ -23,14 +23,16 @@ const Wrapper = styled.View`
 type PropsType = {
   theme: ThemeType,
   city: string,
+  currentLanguage: string,
   languages: Array<LanguageModel>,
-  changeLanguage: (city: string, language: string) => void,
+  changeLanguage: (city: string, newLanguage: string) => void,
```
Schedulaar: Why do we need `city` for changing the language?

src/modules/i18n/LanguageAsyncStorage.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  loadLanguage = (): Promise<?string> => {
+    return this.asyncStorage.getItem(LANGUAGE_KEY)
+  }
+
+  setLanguage = async (language: string) => {
+    await this.asyncStorage.setItem(LANGUAGE_KEY, language)
+  }
+}
```
Schedulaar: We might add `contentLanguage` (as well as `currentCity` later) to the SettingsAsyncStorage to access all stuff in the async storage the same way :thinking:

src/modules/layout/components/Header.js
klinzo: I don't know at all why I added these lines... I just removed them again and everything is fine....

src/modules/i18n/LanguageAsyncStorage.js
klinzo: I thought about that as well, but imo it's clearer to split it. That way, everything that gets displayed in the settings belongs to AppSettings whereas localization stuff has its separate place (I'll rename the LanguageAsyncStorage to Localization)

src/modules/common/components/LanguageNotAvailablePage.js
klinzo: Because if the language is not available we need to do a `navigateToXY` since there is no route mapping in the category state yet and a city is required for it.

src/modules/i18n/LanguageAsyncStorage.js
klinzo: I talked with @maxammann about this at the conference and we agreed that it's fine. But I'm also not really sure.

src/modules/endpoint/reducers/__tests__/morphContentLanguage.spec.js
klinzo: @maxammann what do you propose?

maxammann: Hm yeah you are right, lets go with initialize

src/modules/i18n/LanguageAsyncStorage.js
maxammann: This could interfere with settings:

https://github.com/Integreat/integreat-react-native-app/blob/46edaca973393a32fc52a626341d50e9299b3d30/src/modules/settings/AppSettings.js#L46

I think it is better to put the settings into a `settings` sub-object.

klinzo: Isn't this an unnecessary overhead? There won't be more localization props than `contentLanguage` and `selectedCity`, and those won't be in the settings. So I would rather just  add a comment to mind the other AsyncStorage.

maxammann: As discusses in call: Move to the AppSettings

PR #88, opened 2019-07-01
NATIVE-164: Persist city selection

klinzo:


src/routes/landing/components/Landing.js
```diff
@@ -25,12 +26,35 @@ export type PropsType = {
...
+    const { language, navigateToDashboard } = this.props
+    const localizationSettings = new LocalizationSettings()
+
+    const selectedCity = await localizationSettings.loadSelectedCity()
+
+    if (selectedCity) {
+      navigateToDashboard(selectedCity, language)
+    } else {
+      this.setState({localizationLoaded: true})
+    }
```
Schedulaar: I think this switch should be already done in the Navigator to avoid jumpy screens at the start of the app and not even navigate to the LandingPage first

klinzo: Done. It adds a bit more complexity and there is still kind of a jumpy screen, but should be ok.

src/modules/app/components/Navigator.js
```diff
@@ -74,12 +78,85 @@ export const AppStack = createStackNavigator(
...
+      throw new Error('ContentLanguage must be set!')
+    } else if (!this.navigator) {
+      throw new Error('Ref must not be null!')
+    }
+
+    if (selectedCity) {
+      this.navigateToDashboard(selectedCity, contentLanguage)
+    } else {
+      // $FlowFixMe dispatch is missing in type
+      this.navigator.dispatch(NavigationActions.navigate({
```
Schedulaar: Can't you do sth like `this.navigator.navigate('Landing')`?

src/routes/landing/containers/LandingContainer.js
```diff
@@ -48,28 +47,23 @@ const mapStateToProps = (state: StateType): StatePropsType => {
...
   return {
-    fetchCities: () => dispatch({
-      type: 'FETCH_CITIES'
-    }),
     navigateToDashboard: (cityCode: string, language: string) => {
       const path = `/${cityCode}/${language}`
       const key: string = generateKey()
 
-      const action: NavigationReplaceAction = StackActions.replace({
+      const action = NavigationActions.navigate({
```
Schedulaar: Revert this change; We need to set newKey here and reset the stack:

We need react-navigation to give this route the very same key that we generate above. (Otherwise there's some strange default key, which messes things up)

src/modules/app/components/Navigator.js
```diff
@@ -74,12 +78,85 @@ export const AppStack = createStackNavigator(
...
+      params: {
+        cityCode,
+        key,
+        sharePath: path,
+        onRouteClose: () => this.props.clearCategory(key)
+      }
+    })
+
+    // $FlowFixMe dispatch is missing in type, https://github.com/react-navigation/react-navigation/issues/3842
+    this.navigator.dispatch(NavigationActions.navigate({
```
Schedulaar: I think we should not avoid a direct `dispatch` where possible...
```js
this.navigator.navigate({
  routeName: 'App',
  action: navigateToDashboard
})
```
should do the trick

src/modules/app/components/Navigator.js
```diff
@@ -74,12 +78,85 @@ export const AppStack = createStackNavigator(
...
+        routeName: 'Landing'
+      }))
+    }
+  }
+
+  navigateToDashboard (cityCode: string, language: string) {
+    const path = `/${cityCode}/${language}`
+    const key = generateKey()
+
+    const navigateToDashboard = NavigationActions.navigate({
```
Schedulaar: I guess we need to do that with StackReplace action (as it was done in the Landing Page)

src/modules/app/components/Navigator.js
klinzo: See https://reactnavigation.org/docs/en/app-containers.html#docsNav . `navigator.navigate()` is not working, I've tried it.

src/modules/app/components/Navigator.js
```diff
@@ -74,12 +78,85 @@ export const AppStack = createStackNavigator(
...
+  localizationSettings: LocalizationSettings
+  navigator: ?NavigationContainer<NavigationState, {}, {}>
+
+  constructor (props: PropsType) {
+    super(props)
+    this.localizationSettings = new LocalizationSettings()
+  }
+
+  componentWillMount () {
+    this.props.fetchCities()
```
Schedulaar: Why do we need to fetchCities here?

src/modules/app/components/Navigator.js
```diff
@@ -74,12 +78,85 @@ export const AppStack = createStackNavigator(
...
+    this.navigator.dispatch(NavigationActions.navigate({
+      routeName: 'App',
+      action: navigateToDashboard
+    }))
+
+    this.props.fetchCategory(cityCode, language, key)
+  }
+
+  render () {
+    // eslint-disable-next-line react/jsx-no-bind
```
Schedulaar: I think we should not disable this error here: Every time render is called, the ref prop will always get updated, because the arrow function is generated in every `render` iteration

src/routes/landing/containers/LandingContainer.js
klinzo: Very strange, but okay :D Thanks :)

src/modules/app/components/Navigator.js
klinzo: See above.

src/modules/app/components/Navigator.js
klinzo: Because the fetch is otherwise never called, this was done in the Landing page before but since we skip the Landing page if there is a selected city, the fetch will not happen. We could of course also do this in `loadCityContent`

klinzo: But then we would have to call it at two places.

klinzo:
@Schedulaar how would injecting an initial route config solve the problem that we don't need a Loading screen? The loading route is for the time it takes to decide whether there is a selected city or not. If we set an initial route config, be have the problem with jumpy screens again. Am I understanding something wrong here?

src/modules/app/StateType.js
```diff
@@ -10,7 +10,7 @@ import {
   LanguageModel
 } from '@integreat-app/integreat-api-client'
 import Moment from 'moment'
-import { DEFAULT_LANGUAGE } from '../i18n/components/I18nProvider'
```
maxammann: I would prefere to stick to the i18n name. It is the default for basically every software project.

Also we should try not to rename files if not really needed. (Because of merge conflicts etc.)

src/modules/app/components/Navigator.js
```diff
@@ -74,12 +79,86 @@ export const AppStack = createStackNavigator(
...
+|}
+
+class Navigator extends React.Component<PropsType> {
+  localizationSettings: LocalizationSettings
+  navigator: {current: null | React$ElementRef<NavigationContainer<NavigationState, {}, {}>>}
+
+  constructor (props: PropsType) {
+    super(props)
+    this.localizationSettings = new LocalizationSettings()
+    this.navigator = React.createRef()
```
maxammann: I'm pretty sure we can move the `navigator` reference to a child of this component.

Can the function `loadSelectedCity` be called in the `componentWillMount` of a child component? Then you can use the `withNavigation` hoc to get a reference to the navigator

src/modules/localization/LocalizationSettings.js
```diff
@@ -18,6 +19,18 @@ class LanguageAsyncStorage {
...
+    await this.asyncStorage.setItem(SELECTED_CITY, city)
+  }
+
+  clearSelectedCity = async () => {
+    await this.asyncStorage.removeItem(SELECTED_CITY)
+  }
 }
 
-export default LanguageAsyncStorage
+export default LocalizationSettings
```
maxammann: Should be moved to AppSettings as discussed in call.

src/modules/app/components/Navigator.js
klinzo: Could you elaborate further what you mean by that? If I understand you correctly, you propose to introduce a child component to handle the navigating as soon as it is known whether a city is selected or not.
Afaics you can't do any navigating outside the `AppContainer` without using a `ref`. I think `withNavigation` would only work inside a child component of the `AppContainer`, since the `navigation` is only defined there.

maxammann: Yes that's what I meant. Not sure whether that works in this case though. The idea would be, as you said, to create a child component.

Afaik refs shouldn't be needed in 99% of the cases.

src/modules/app/StoreActionType.js
```diff
@@ -123,6 +123,10 @@ export type ClearCityContentActionType = {|
   type: 'CLEAR_CITY_CONTENT'
 |}
 
+export type ClearCityActionType = {|
```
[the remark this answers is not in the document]
klinzo: We need two actions because we need to call asynchronous methods (`clearCitySelection`) and modify the state. So one is a saga to call the async method and one a normal redux reducer.

src/modules/app/components/Navigator.js
```diff
@@ -74,12 +79,86 @@ export const AppStack = createStackNavigator(
...
+    this.navigator.current.dispatch(NavigationActions.navigate({
+      routeName: 'App',
+      action: navigateToDashboard
+    }))
+
+    this.props.fetchCategory(cityCode, language, key)
+  }
+
+  render () {
+    return <AppContainer ref={this.navigator} />
```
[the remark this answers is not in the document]
klinzo: Nope, just a white screen shows up. Should we create a task to show the spalsh-screen here?

PR #95, opened 2019-07-08
NATIVE-299: Fix language switching while loading

klinzo:


src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -47,6 +47,21 @@ function * fetchCategory (dataContainer: DataContainer, action: FetchCategoryAct
...
+    response: call(fetchCategory, dataContainer, action),
+    cancel: take('SWITCH_CONTENT_LANGUAGE')
+  })
+
+  if (cancel) {
+    const newFetchCategory = action
+    const newLanguage = cancel.params.newLanguage
+    newFetchCategory.params.language = newLanguage
+    newFetchCategory.params.path = `/${action.params.city}/${newLanguage}`
+    yield put(newFetchCategory)
```
maxammann: Create a new action here and copy the relevant stuff over. Immutable data ftw :)

src/modules/fetcher/FetcherModule.js
```diff
@@ -10,19 +10,20 @@ export type FetchResultType = { [path: string]: {| lastUpdate: string, url: stri
...
+  // static currentlyFetching = false
 
   async fetchAsync (
     targetFilePaths: TargetFilePathsType,
     progress: ProgressCallbackType
   ): Promise<FetchResultType> {
-    if (FetcherModule.currentlyFetching) {
-      return Promise.reject(new Error('Already fetching!'))
-    }
+    // if (FetcherModule.currentlyFetching) {
```
maxammann: Test whether this works on iOS and Android.

If this works on both platforms this would make the task NATIVE-264 easier. ALso note it there.

klinzo: I don't have a possibility to test this on iOS, but this is only temporary and should be fixed in NATIVE-264, so I don't see exactly why this should be done here.

maxammann: Hm I'm not sure whether we should trade one bug for an other one. I think this should be easily testable on browserstack, right?

Else I have a spare iphone for such tasks now around.

klinzo: So I don't really understand your point. By doing this, we don't trade a bug for another one since the `already fetching` bug is probably already existing and if not, this isn't a bug at all. Apart from that, there is already an issue to check if the problem still exists and fix it if so.
If you like though, I could also revert the `FetcherModule`, so everything stays the same.

maxammann: My point is that afaik removing the `throw` does not change anything. It just adds a very weird crash. If I remember correctly the app just crashes without any error.

I'm not really sure what the error is in this task because it is not documented in the issue. If the error is "Already fetching!" then the changes above do not fix it. If there is an another error shown then the above is probably a fix.

So if the latter is true then just revert the changes in FetcherModule and we will address this in `NATIVE-264`.

klinzo: Ah yeah ok now I get it. Yeah I'll just revert the FetcherModule as it doesn't make sense to do it in this task.

src/modules/endpoint/reducers/morphContentLanguage.js
```diff
@@ -7,13 +7,13 @@ import type { MorphContentLanguageActionType } from '../../app/StoreActionType'
 import CategoriesMapModel from '@integreat-app/integreat-api-client/models/CategoriesMapModel'
 import forEachTreeNode from '../../common/forEachTreeNode'
 
-const categoryRouteTranslator = (newCategoriesMap: CategoriesMapModel, newLanguage: string) =>
+const categoryRouteTranslator = (newCategoriesMap: ?CategoriesMapModel, newLanguage: string) =>
```
Schedulaar: this should never be undefined or null

klinzo:
@maxammann ping

src/modules/endpoint/reducers/morphContentLanguage.js
```diff
@@ -27,6 +27,7 @@ const categoryRouteTranslator = (newCategoriesMap: CategoriesMapModel, newLangua
     const resultModels = {}
     const resultChildren = {}
 
+    // $FlowFixMe flow is fucking stupid here and complains that getChildren is missing in null or undefined
```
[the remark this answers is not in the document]
maxammann: Censoring! I'm actually pro using rude curse words :P Let the people express them selves during development!

https://www.vidarholen.net/contents/wordcount/

(I'm serious about this ;))

PR #98, opened 2019-07-08
NATIVE-31: Add sharing for PDFs/Images

maxammann:
This PR adds sharing feature for files.
Added default message if share fails.

Screenshots:
image
image

Fixes: [NATIVE-31](https://issues.integreat-app.de/browse/NATIVE-31)


maxammann:
Missing in this PR: 
* Take care about 'subject' and 'title' - We probably want to ignore these to be consistent across platforms

PR #99, opened 2019-07-08
NATIVE-273: Swap color of navigation tiles

Schedulaar:


src/modules/common/components/NavigationTile.js
```diff
@@ -22,9 +22,15 @@ const Circle = styled(View)`
   border-radius: ${ICON_SIZE}px;
   height: ${ICON_SIZE};
   width: ${ICON_SIZE};
-  background-color: ${props => props.theme.colors.themeColor};
+  background-color: ${props => props.theme.colors.backgroundColor};
   align-items: center;
   justify-content: center;
+  elevation: 2;
```
maxammann: Is this elevation equivalent to the shadow below for iOS? If yes does this fix this issue? https://issues.integreat-app.de/browse/NATIVE-48

Schedulaar: Now it is using https://ethercreative.github.io/react-native-shadow-generator/

PR #107, opened 2019-07-16
NATIVE-191: Cleanup i18n provider

klinzo:


src/modules/app/StateType.js
```diff
@@ -86,7 +86,6 @@ export const defaultCityContentState = null
 export type DirectionStateType = 'ltr' | 'rtl'
```
Schedulaar: You might (or might not) remove this for now as well...

PR #108, opened 2019-07-17
NATIVE-8: Update documentation

klinzo:
Also: NATIVE-274: Add note about renaming from `news` to `events`


docs/14-i18n.md
```diff
@@ -0,0 +1,34 @@
...
+
+The app has two languages:
+* **UI language**
+* **Content language**
+
+## UI language
+
+The UI language is always the primary system language and cannot be changed.
+
+All UI and navigation elements are using the UI language and the corresponding layout direction, i.e. LTR or RTL.
```
maxammann: We should decide here for a naming of the elements. I think UI elements is fine.

docs/14-i18n.md
```diff
@@ -0,0 +1,34 @@
...
+
+The content language is changeable in the app and is set to the UI language at the first app start.
+
+All content elements are using the content language and the corresponding layout direction, i.e. LTR or RTL.
+Content elements are all elements that are NOT UI elements, especially
+* Webview content
+* Categories, Events, Offers tiles and lists
+* LastUpdate
+
+## WIP: To discuss
```
maxammann: I would go for this rule:
_If the element is important for a user who is shown the app during counseling, then it is content language. _

Search, Feedback, Language change and Landing page are not important for such users.

maxammann: We should also finish this here as there shouldn't be an other task imo

klinzo: What do you mean by `finish this here`? Just  to decide whether the items in `To discuss` are belonging to content or ui? That was the plan!

klinzo: And I like your rule. @Schedulaar what do you think?

docs/14-i18n.md
Schedulaar: Looking good to me.

PR #109, opened 2019-07-21
NATIVE-44: Add retry button to error screens

Schedulaar:
Unfortunately, I had to overcome some architectural problems with displaying the errors; that's why this PR is blown up so much.

Also there's one $FlopFixMe I had to add, I don't know how to solve this one :/ If you don't  know either, I'd suggest to leave it there for now


src/modules/error/hocs/withPayloadProvider.js
```diff
@@ -0,0 +1,92 @@
...
+  RouteNotInitializedType
+  | LoadingType
+  | ErrorType<R>
+  | LanguageNotAvailableType<R>
+  | SuccessType<$Diff<S, { dispatch: Dispatch<StoreActionType> }>, R>
+
+export type PropsType<S: { dispatch: Dispatch<StoreActionType> }, R> = {|
+  ...StatusPropsType<S, R>,
+  dispatch: Dispatch<StoreActionType>
+|}
```
klinzo: Well done, I wasn't able to find this solution... :+1:

src/modules/error/hocs/withPayloadProvider.js
```diff
@@ -0,0 +1,92 @@
...
+          </ScrollView>
+        } else if (props.status === 'languageNotAvailable') {
+          return <LanguageNotAvailableContainer languages={props.availableLanguages}
+                                                changeLanguage={this.changeUnavailableLanguage} />
+        } else if (props.status === 'loading') {
+          return <ScrollView refreshControl={<RefreshControl refreshing />} contentContainerStyle={{ flexGrow: 1 }} />
+        } else { // props.status === 'success'
+          return <ScrollView refreshControl={<RefreshControl onRefresh={this.refresh} refreshing={false} />}
+                             contentContainerStyle={{ flexGrow: 1 }}>
+            <Component {...props.innerProps} dispatch={props.dispatch} />
```
klinzo: Do we need to pass `dispatch`?

src/modules/app/containers/NavigatorContainer.js
```diff
@@ -3,15 +3,24 @@
...
+  clearCategory: (key: string) => void,
+  fetchCities: () => void
 |}
 
 type PropsType = DispatchPropsType
 
-const mapDispatchToProps = (dispatch: Dispatch<*>): DispatchPropsType => ({
+const mapDispatchToProps = (dispatch: Dispatch<StoreActionType>): DispatchPropsType => ({
+  setContentLanguage: (language: string) => {
+    dispatch({
```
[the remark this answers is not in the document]
Schedulaar: This is not necessary, since I already typed the dispatch function with `Dispatch<StoreActionType>` which in turn only allows an action of this type to be dispatched.

src/modules/app/containers/NavigatorContainer.js
```diff
@@ -30,6 +39,9 @@ const mapDispatchToProps = (dispatch: Dispatch<*>): DispatchPropsType => ({
   },
   clearCategory: (key: string) => {
     dispatch({ type: 'CLEAR_CATEGORY', params: { key } })
+  },
+  fetchCities: () => {
+    dispatch({ type: 'FETCH_CITIES' })
```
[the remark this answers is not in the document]
Schedulaar: same answer as above

src/modules/app/createNavigateToEvent.js
```diff
@@ -6,15 +6,14 @@ import type { NavigationScreenProp } from 'react-navigation'
 import { generateKey } from './generateRouteKey'
 
 export type NavigateToEventParamsType =
-  {|cityCode: string, language: string, path?: string, key?: string, forceUpdate?: boolean|}
+  {|cityCode: string, language: string, path?: ?string, key?: string, forceUpdate?: boolean|}
```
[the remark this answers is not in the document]
Schedulaar: passing `path === null` means navigating to the events-list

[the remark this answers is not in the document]
klinzo: Or undefined

src/locales.json
```diff
@@ -220,6 +220,7 @@
         "disclaimer": "Ein Impressum ist leider für diese Stadt nicht verfügbar.",
         "pageTitle": "Seite nicht gefunden"
       },
+      "tryAgain": "Nochmal Versuchen",
```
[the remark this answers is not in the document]
maxammann: I vote for just "Nommal!"

src/modules/error/hocs/withPayloadProvider.js
Schedulaar: The alternative would be to connect the `Component` which would result in two `connect`s in all of the containers (CategoriesContainer, DashboardContainer, etc.)
Since I needed the dispatch in all of the cases, I decided to pass it here...

src/modules/app/components/Navigator.js
```diff
@@ -62,7 +67,9 @@ export const AppStack = createStackNavigator(
     [WOHNEN_ROUTE]: createNavigationRouteConfig(WohnenExtraContainer, defaultHeader),
     [SPRUNGBRETT_ROUTE]: createNavigationRouteConfig(SprungbrettExtraContainer, defaultHeader),
     [EXTERNAL_EXTRA_ROUTE]: createNavigationRouteConfig(ExternalExtraContainer, defaultHeader),
-    'Events': createNavigationRouteConfig(EventsContainer, defaultHeader),
+    'Events': createNavigationRouteConfig( // $FlowFixMe
```
Schedulaar: If you have any idea, on how to solve this, don't hesitate to suggest it :D
I cannot see any significant difference of EventsContainer to DashboardContainer or CategoriesContainer why Flow would throw an error here, but not at the other screens... :/

src/modules/endpoint/sagas/watchContentLanguageSwitch.js
```diff
@@ -23,6 +24,10 @@ function * switchContentLanguage (dataContainer: DataContainer, action: SwitchCo
     }
     yield put(setContentLanguage)
 
+    const city = yield select((state: StateType) => state.cityContent && state.cityContent.city)
```
klinzo: Why don't you just pass `city` as action param?

src/routes/dashboard/components/Dashboard.js
```diff
@@ -76,41 +70,18 @@ class Dashboard extends React.Component<PropsType> {
...
-    }
-  }
-
   render () {
     const {
       cities, stateView, theme, resourceCache, navigateToIntegreatUrl, language, cityCode, navigateToCategory,
       navigation, t
     } = this.props
 
-    if (!stateView || !cities || !resourceCache || !cityCode || !language) {
```
klinzo: :heart_eyes:

src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -4,94 +4,139 @@ import { connect } from 'react-redux'
...
+  const { cityCode, language, path, navigation } = refreshProps
+  const navigateToCategories = createNavigateToCategory('Categories', dispatch, navigation)
+  navigateToCategories({
+    cityCode, language, path, forceUpdate: true, key: navigation.getParam('key')
+  })
+}
+
+class CategoriesContainer extends React.Component<ContainerPropsType> {
+  render () {
+    const { cities, language, stateView, cityCode, navigation, resourceCache, dispatch } = this.props
```
klinzo:
```suggestion
    const {dispatch, navigation, ...rest} = this.props
```

src/modules/app/containers/NavigatorContainer.js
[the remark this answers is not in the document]
klinzo: :astonished:

src/modules/app/createNavigateToEvent.js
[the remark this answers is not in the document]
klinzo: So my question was meant as: Do we need both?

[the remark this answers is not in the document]
Schedulaar: Ah no we don't need both question marks, I removed one...

src/modules/endpoint/sagas/watchContentLanguageSwitch.js
Schedulaar: Well, it's not necessary to have a city; actions should be as atomic as they can be imo...

src/routes/categories/containers/CategoriesContainer.js
Schedulaar: Done.

src/modules/endpoint/sagas/watchContentLanguageSwitch.js
klinzo: I don't get that... what is the difference of passing `city: ?string` as action param or reading it from the state by `select`? The  docs regarding `select` are saying the following: *A saga should preferably depend only on its own internal control state when possible* and I kind of agree with that here because I don't really see the need why we need to do this here. If there is a reason to do this here, it probably makes sense in our other sagas as well.

src/modules/app/components/Navigator.js
klinzo: I don't have a clue, sorry. But if nobody knows a solution, it's not worth the time to look for a solution.

maxammann: Yeah lets not waste time on this. Maybe just add a comment to the FlowFixMe that we do not know why it happens. Also a task would be good maybe as it is in a central component of the app.

src/modules/app/createNavigateToIntegreatUrl.js
```diff
@@ -22,7 +22,7 @@ export default (dispatch: Dispatch<StoreActionType>, navigation: NavigationScree
         createNavigateToEvent(dispatch, navigation)({ cityCode, language: newLanguage, path: pathname })
       } else {
         // '/augsburg/de/events'
-        createNavigateToEvent(dispatch, navigation)({ cityCode, language: newLanguage })
+        createNavigateToEvent(dispatch, navigation)({ cityCode, language: newLanguage, path: null })
```
maxammann: Minor: `undefined` would be more clear imo. Maybe we should find a rule of thumb of null vs undefined.

src/modules/endpoint/reducers/citiesReducer.js
```diff
@@ -8,6 +8,10 @@ export default (
   state: CitiesStateType = defaultCitiesState, action: CitiesActionType
 ): CitiesStateType => {
   switch (action.type) {
+    case 'FETCH_CITIES':
+      return {
+        models: null
```
maxammann: I have a discussion here: So if we discard data right when we start loading then it will be lost in case of an error.
I think this is oke but it would be better if the previous data is shown in that case. What do you think?

src/modules/endpoint/reducers/cityContentReducer.js
```diff
@@ -6,51 +6,86 @@ import morphContentLanguage from './morphContentLanguage'
...
-  if (!state) {
-    return defaultCityContentState
-  }
-
   switch (action.type) {
     case 'SWITCH_CONTENT_LANGUAGE':
-      return { ...state, switchingLanguage: true }
+      return state === null ? null : { ...state, switchingLanguage: true }
+    case 'PUSH_LANGUAGES':
+      if (state === null) {
```
maxammann: There is quite some duplicate code about checking for `state === null`.

We could split this split case (first part needs `state !== null`) such that each part is in a separate function. Then we can check at the start of the function for the state and throw an error.

src/modules/endpoint/sagas/watchContentLanguageSwitch.js
maxammann: I'm also for passing this as a param if possible. It is way more difficult to tell which value `cityContent.city` has at that exact point because it depends on previously `putted` actions.

maxammann: Keeping functions and sagas pure (as they tell in the docu) is generally a good style.

src/modules/error/hocs/withPayloadProvider.js
```diff
@@ -0,0 +1,92 @@
+// @flow
```
maxammann: Hell :imp:  of a HOC but yeah probably no easier way right now :D

src/modules/endpoint/reducers/citiesReducer.js
Schedulaar: Well, it should not be lost completely, since the datacontainer / database might still hold the old data... But yeah, I see your problem, that we might want to display data when a refresh failed (if there's offline data available). So we might refill the state with the old data if available (and for example notify the user of the error by using a Toast triggered in a Saga?)
Of course this is not part of this task, although we definitely need to discuss on it...

src/modules/endpoint/sagas/watchContentLanguageSwitch.js
Schedulaar: Yeah, I might change it back... I'm still a little bit struggling with the asynchronous nature of sagas though (in theory it could happen that the `city` in the state is different to when the saga was fired (due to actions that dispatched while the saga was running), which might lead to other problems, if the saga dispatches further actions to the state assuming a wrong `city` 🤔

src/modules/error/hocs/withPayloadProvider.js
Schedulaar: I like it though ^^ Streamlines the other components quite a lot

src/modules/app/createNavigateToIntegreatUrl.js
Schedulaar: Hmm... Yeah, I don't really like `undefined` though... Not sure what's best here; maybe avoiding it by using another field to distinguish `Events` and `Event`...
I'll leave it like this for now.

src/modules/endpoint/reducers/citiesReducer.js
Schedulaar: I created a task for this: https://issues.integreat-app.de/browse/NATIVE-334

src/modules/endpoint/sagas/watchContentLanguageSwitch.js
Schedulaar: Changed it back

src/modules/endpoint/reducers/cityContentReducer.js
Schedulaar: I will leave it like this, since I think we should check this for each action, to make it more explicit...

PR #110, opened 2019-07-22
NATIVE-318: Handle rtl

klinzo:


src/modules/common/renderHtml.js
```diff
@@ -53,7 +53,7 @@ const renderJS = (files: FileCacheStateType) => `
 })();
 `
 
-export default (html: string, files: FileCacheStateType, theme: ThemeType) => {
+export default (html: string, files: FileCacheStateType, theme: ThemeType, direction: string) => {
```
Schedulaar:
```suggestion
export default (html: string, files: FileCacheStateType, theme: ThemeType, direction: 'ltr' | 'rtl') => {
```

src/modules/common/renderHtml.js
```diff
@@ -152,7 +152,7 @@ export default (html: string, files: FileCacheStateType, theme: ThemeType) => {
     }
   </style>
 </head>
-<body>
+<body dir="${direction}">
```
maxammann: Is this really required here? What happens to arabic/german text when this is not set?

As far as I understand unicode this should not be needed. Maybe I'm wrong. If not then propably an other fix is needed.

Schedulaar: This is absolutely necessary here (for example for setting the direction for images, tables, etc.). We do that in the webapp as well.

PR #112, opened 2019-07-22
NATIVE-33: Change header

klinzo:


src/modules/categories/components/Categories.js
```diff
@@ -159,7 +159,7 @@ class Categories extends React.Component<PropsType> {
       return <SpaceBetween>
         <View>
           <Tiles tiles={this.getTileModels(children)}
-                 title={CityModel.findCityName(cities, category.title)}
+                 title={null}
```
maxammann: I think it would make more sense to use undefined here and just remove the `tile=null`

src/modules/layout/components/Header.js
```diff
@@ -130,14 +142,15 @@ class Header extends React.PureComponent<PropsType> {
   }
 
   render () {
-    const { navigation, t, theme } = this.props
+    const { cityModel, navigation, t, theme } = this.props
     const sharePath = navigation.getParam('sharePath')
+    const cityDisplayName = cityModel && `${cityModel.sortingName}${cityModel.prefix ? ` (${cityModel.prefix})` : ''}`
```
maxammann: That line is a little bit too complicated imo. 
Best way would be to create a component called `HeaderText` which accepts a cityModel as prop.
Easiest change would be to create a function which renders a HeaderText

PR #113, opened 2019-08-01
NATIVE-170: Mapping city model

Simple-codinger:


src/modules/endpoint/DatabaseConnector.js
```diff
@@ -52,6 +53,15 @@ type ContentEventJsonType = {|
   |}
 |}
 
+type ContentCityJsonType = {|
+    name: string,
+    live: boolean,
+    code: string,
+    extrasEnabled: boolean,
+    eventsEnabled: boolean,
+    sortingName: string
```
klinzo: `prefix` is missing here, was added some time ago

src/modules/endpoint/DatabaseConnector.js
```diff
@@ -87,6 +98,10 @@ class DatabaseConnector {
   }
 
   getMetaCitiesPath (): string {
+    return `${CACHE_DIR_PATH}/metaCities.json`
```
maxammann: By following the current naming schema this should be:
```suggestion
    return `${CACHE_DIR_PATH}/cities-meta.json`
```

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -100,8 +100,7 @@ class DefaultDataContainer implements DataContainer {
     await cache.cache(categories, context)
   }
 
-  setCities = (cities: Array<CityModel>) => {
-    // TODO: Offline available cities will be persisted in NATIVE-175. For now switching cities when offline is not possible.
+  setCities = async (cities: Array<CityModel>) => {
     this._cities = cities
     return Promise.resolve()
```
maxammann: I think this should write the cities.json.

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -100,10 +100,9 @@ class DefaultDataContainer implements DataContainer {
     await cache.cache(categories, context)
   }
 
-  setCities = (cities: Array<CityModel>) => {
-    // TODO: Offline available cities will be persisted in NATIVE-175. For now switching cities when offline is not possible.
+  setCities = async (cities: Array<CityModel>) => {
     this._cities = cities
-    return Promise.resolve()
+    await this._databaseConnector.storeCities(cities)
```
maxammann: I'm not 100% sure I think you have to return the promise. Else it is just resolved as soon as the function returns.
```suggestion
    return this._databaseConnector.storeCities(cities)
```

src/modules/endpoint/sagas/watchFetchCities.js
```diff
@@ -14,11 +14,14 @@ import { baseUrl } from '../constants'
...
 function * fetchCities (dataContainer: DataContainer): Saga<void> {
   try {
-    const payload: Payload<Array<CityModel>> = yield call(() => request(createCitiesEndpoint(baseUrl)))
-
-    const cities: Array<CityModel> = payload.data
-
-    yield call(dataContainer.setCities, cities)
+    let cities: Array<CityModel>
+    if (yield call(() => dataContainer.citiesAvailable())) {
+      cities = yield call(() => dataContainer.getCities())
```
Schedulaar: The user should be able to force a refresh (using pull-to-refresh) of the cities.
If you want, you can implement that in this task or in another task.
If you choose the latter please create one for that :)

PR #114, opened 2019-08-05
NATIVE-100: Add last adjustments

maxammann:


src/modules/endpoint/sagas/loadCategories.js
```diff
@@ -22,8 +22,10 @@ function * loadCategories (
   dataContainer: DataContainer,
   shouldUpdate: boolean
 ): Saga<FetchMapType> {
-  const context = new DatabaseContext(city, language)
-  const categoriesAvailable = yield call({ context: dataContainer, fn: dataContainer.categoriesAvailable }, context)
+  const categoriesAvailable = yield call(
+    (city: string, language: string) => dataContainer.categoriesAvailable(city, language),
```
klinzo: Wouldn't this work?
```suggestion
   yield call(dataContainer.categoriesAvailable, city, language)
```

src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -49,8 +47,8 @@ export default function * loadCityContent (
   console.debug('City content should be refreshed: ', shouldUpdate)
 
   if (criterion.shouldLoadLanguages()) {
-    yield call(loadLanguages, context, dataContainer, shouldUpdate)
-    const languages = yield call(dataContainer.getLanguages, context)
+    yield call(loadLanguages, newCity, dataContainer, shouldUpdate)
+    const languages = yield call(dataContainer.getLanguages, newCity, newLanguage)
```
klinzo: `newLanguge` shouldn't be required I guess

src/modules/endpoint/sagas/loadEvents.js
```diff
@@ -19,8 +18,10 @@ function * fetchEvents (city: string, language: string): Saga<Array<EventModel>>
 function * loadEvents (
   city: string, language: string, dataContainer: DataContainer, shouldUpdate: boolean
 ): Saga<FetchMapType> {
-  const context = new DatabaseContext(city, language)
-  const eventsAvailable = yield call({ context: dataContainer, fn: dataContainer.eventsAvailable }, context)
+  const eventsAvailable = yield call(
+    (city: string, language: string) => dataContainer.eventsAvailable(city, language),
```
klinzo: Same as above

src/modules/endpoint/sagas/loadLanguages.js
```diff
@@ -6,17 +6,22 @@ import { createLanguagesEndpoint, LanguageModel, Payload } from '@integreat-app/
...
-import DatabaseContext from '../DatabaseContext'
 
-export default function * loadLanguages (context: DatabaseContext, dataContainer: DataContainer, shouldUpdate: boolean): Saga<void> {
-  const languagesAvailable = yield call({ context: dataContainer, fn: dataContainer.languagesAvailable }, context)
+export default function * loadLanguages (
+  city: string,
+  dataContainer: DataContainer, shouldUpdate: boolean
+): Saga<void> {
+  const languagesAvailable = yield call(
+    (city: string) => dataContainer.languagesAvailable(city),
```
klinzo: Same again

src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -10,16 +10,15 @@ import type {
...
-import DatabaseContext from '../DatabaseContext'
 import isPeekingRoute from '../selectors/isPeekingRoute'
 
 function * getRootAvailableLanguages (
-  context: DatabaseContext,
+  city: string, language: string,
   loadCriterion: ContentLoadCriterion, dataContainer: DataContainer): Saga<Map<string, string>> {
   if (loadCriterion.shouldLoadLanguages()) {
-    const languages = yield call(dataContainer.getLanguages, context)
+    const languages = yield call(dataContainer.getLanguages, city, language)
```
klinzo: `language` should not be required I guess

src/modules/endpoint/sagas/loadCategories.js
maxammann: No, because categoriesAvailable(...) references `this`

src/modules/endpoint/sagas/loadEvents.js
maxammann: Events require the language.

src/modules/endpoint/sagas/loadLanguages.js
maxammann: only city is required here

src/modules/endpoint/sagas/loadCategories.js
klinzo: Shouldn't we just use arrow functions then? Maybe we should find a rule of thumb when to use which (for the sake of consistency)

src/modules/endpoint/sagas/loadEvents.js
klinzo: I meant the ` yield call(dataContainer.categoriesAvailable, city, language)` proposal, but apparently the same answer.

src/modules/endpoint/sagas/loadLanguages.js
klinzo: I meant the ` yield call(dataContainer.categoriesAvailable, city, language)` proposal, but apparently the same answer.

src/modules/endpoint/sagas/loadCategories.js
maxammann: It depends on the function you are calling. If you call a pure and static function you do not need lambdas. If you are calling a nonpure function on an instance you need a lambda or that other syntax.

PR #117, opened 2019-08-12
NATIVE-114: Test app module

Schedulaar:


src/modules/app/generateRouteKey.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 
 // This code is adopted from: https://github.com/react-navigation/react-navigation-core/blob/c85a22d78a50e853632a11036ce562e4e1ecb523/src/routers/KeyGenerator.ts
-const uniqueBaseId: string = `category-route-id-${Date.now()}`
+const uniqueBaseId: string = `route-id-${Date.now()}`
```
klinzo: I think there is an issue for this, could be rejected now.

klinzo: https://issues.integreat-app.de/browse/NATIVE-180

Schedulaar: Thanks. I rejected it.

src/modules/app/SentryIntegration.js
```diff
@@ -24,7 +24,7 @@ export default class SentryIntegration {
       return
     }
 
-    this.sentry.install()
+    await this.sentry.install()
```
maxammann: Not sure whether we need/want to wait on this if it is not needed. The sentry API behaves a little bit unexpected.

Official docs do not await it: https://docs.sentry.io/clients/react-native/#configure

src/modules/app/__tests__/CategoriesRouteStateView.spec.js
```diff
@@ -0,0 +1,68 @@
...
+  })
+
+  it('should return models of children when calling children', () => {
+    const stateView = new CategoriesRouteStateView('/augsburg/de', models, children)
+    expect(stateView.children()).toEqual([models['/augsburg/de/erste-hilfe']])
+  })
+
+  it('should enable stepping into a child category', () => {
+    const stateView = new CategoriesRouteStateView('/augsburg/de', models, children)
+    const child = stateView.stepInto('/augsburg/de/erste-hilfe')
```
maxammann: Maybe it would be beneficial to test the behavior if the path (paramof stepInto) is not available.

src/modules/app/__tests__/createNavigateToIntegreatUrl.spec.js
```diff
@@ -0,0 +1,110 @@
...
+    navigateToLanding: jest.fn(),
+    navigateToEvent: jest.fn(),
+    navigateToCategory: jest.fn(),
+    navigateToDashboard: jest.fn()
+  }
+  const navigateToIntegreatUrl = createNavigateToIntegreatUrl(mocks)
+  return { navigateToIntegreatUrl, mocks }
+}
+
+const expectExactlyOneHit = (mocks: {
```
maxammann: :+1:

src/routes/events/containers/EventsContainer.js
```diff
@@ -126,7 +126,7 @@ class EventsContainer extends React.Component<ContainerPropsType> {
 const refresh = (refreshProps: RefreshPropsType, dispatch: Dispatch<StoreActionType>) => {
   const { navigation, cityCode, language, path } = refreshProps
   const navigateToEvent = createNavigateToEvent(dispatch, navigation)
-  navigateToEvent({ cityCode, language, path, forceUpdate: true, key: navigation.getParam('key') })
+  navigateToEvent({ cityCode, language, path, forceUpdate: true, key: navigation.state.key })
```
maxammann: Not sure about this change, but if it works it is probably oke :)

src/modules/app/SentryIntegration.js
Schedulaar: Well otherwise it's not possible catch any errors that happen when installing it (see https://github.com/Integreat/integreat-react-native-app/blob/4ded4229bcd35aaa2949e02bed290f82f03b2e26/src/modules/app/components/LaunchInquiry.js#L84). Also the example you refer to is very minimalistic I guess...

src/routes/events/containers/EventsContainer.js
Schedulaar: Yeah, having the navigation in the navigation params doesn't make a lot of sense, since it's available through the navigation state

src/modules/app/__tests__/CategoriesRouteStateView.spec.js
Schedulaar: Hmm...  I think, `stepInto` should only be called with param `path` if the user of this method already knows, that `path` is in the subtree of the stateView (which is true).
So this use-case should not be part of the tests, but I'll add some documentation.

Schedulaar:
@maxammann I decided to replace the routeStateView with a simpler tree-structure of categories. That way we don't have to unnaturally flatten the CategoriesTree (using forEachTreeNode and a models and children array) just to present a tree-like API in the categoriesRouteStateView. This reduces complexity (and possible mistakes), so please have another look at the last commit and tell me what you think

maxammann:
https://redux.js.org/recipes/structuring-reducers/normalizing-state-shape

Schedulaar:
As I don't want to discuss that in this PR, I reverted the change :)

klinzo:
Ready to merge?

Schedulaar:
LaunchInquiry and Navigator still need to be tested. I'll extract them from this issue and open a new one for them.

PR #118, opened 2019-08-19
NATIVE-339: Add cache for cities

Schedulaar:
I thought the app did not start in the dev branch. That was because there were some local files (cities.json) still present which held the wrong data (as of the city-mapping-task). After deleting the app files  (several times) it finally did work, but I already made some changes, that streamlines the DatabaseConnector, so I thought I'd still opened a PR for it.

### DIFF
--- a/src/modules/endpoint/DefaultDataContainer.js
+++ b/src/modules/endpoint/DefaultDataContainer.js
@@ -11,6 +11,7 @@ import RNFetchBlob from 'rn-fetch-blob'
 import Cache from './Cache'
 
 type CacheType = {
+  cities: Cache<Array<CityModel>>,
   events: Cache<Array<EventModel>>,
   categories: Cache<CategoriesMapModel>,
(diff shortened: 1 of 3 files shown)


src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -106,8 +106,8 @@ class DefaultDataContainer implements DataContainer {
   }
 
   setCities = async (cities: Array<CityModel>) => {
-    this._cities = cities
-    return this._databaseConnector.storeCities(cities)
+    const cache = this.caches.cities
+    await cache.cache(cities, new DatabaseContext())
```
maxammann: Not sure whether I like a DatabaseContext with null values, but I think it is acceptable :)

Schedulaar: You mean with only null values? I mean you introduced language to be nullable, so as we have data that is independent of the city as well, it makes sense to have one with city being null as well :)

PR #120, opened 2019-08-19
NATIVE-325: Test endpoint reducers

Schedulaar:


src/modules/endpoint/reducers/__tests__/__snapshots__/morphContentLanguage.spec.js.snap
```diff
@@ -77,17 +77,18 @@ Object {
...
     },
   },
   "city": "augsburg",
   "eventsRouteMapping": Object {
     "event-route-1": Object {
       "allAvailableLanguages": Map {
-        "de" => "de",
-        "en" => "en",
+        "de" => "/augsburg/de/events",
+        "en" => "/augsburg/en/events",
```
maxammann: :+1:

src/modules/app/StateType.js
```diff
@@ -13,7 +13,7 @@ import { DEFAULT_LANGUAGE } from '../i18n/components/I18nProvider'
 export type PathType = string
 
 export type CategoryRouteConfigType = {|
-  +root: string, // path of the root category
+  +path: string,
```
maxammann: Hm, I favor the name `root`. As it is the rot of a sub-tree of the CategoryMap. Maybe `rootPath` would be oke. Atleast I'm against using more general names like `path`. We already have enough parameters which are called `path`.

maxammann: The nameing should be consistent with `CategoriesRouteStateView` as it corresponds to the state.

Schedulaar: As far as I can see, CategoriesRouteStateView (btw: what's a routeStateView? - some view on the routeState?) doesn't really correspond to the state, since the CategoryRouteStateType holds more information than the CategoriesRouteStateView.

Also in comparison with the events state it makes a lot of sense to use `path` here, since it just corresponds to the actual path that the route should display. `root` is super confusing to me as this suggests that the value says something about the root category, which is not the case...

src/modules/endpoint/reducers/__tests__/cityContentReducer.spec.js
```diff
@@ -0,0 +1,297 @@
...
+    const prevState: CityContentStateType = {
+      city: 'augsburg',
+      categoriesRouteMapping: {},
+      eventsRouteMapping: {},
+      languages: [],
+      resourceCache: {},
+      searchRoute: null,
+      switchingLanguage: false
+    }
+    expect(cityContentReducer(prevState, switchContentLanguageAction)?.switchingLanguage).toBe(true)
```
klinzo: Is the `?` usable now? Nice!

Schedulaar: Yup :unicorn: :rainbow:

src/modules/app/StateType.js
maxammann: You are correct CategoriesRouteStateView is a view on the state. As it is a view it is oke state the state contains more information than the view.

I'm oke with changing it to `path` as long as view and state stay consistent: https://github.com/Integreat/integreat-react-native-app/blob/56845e36c376840b6fd3eb9be011ccf43ec21158/src/modules/app/CategoriesRouteStateView.js#L7

`rawRoot` should be changed to `rawPath`.

PR #121, opened 2019-08-20
NATIVE-325: Add pull to refresh to landing

Schedulaar:
I accidentally started (and finished) working on this, so I thought I'd open a PR although this issue has not been in this sprint.

### DIFF
--- a/src/modules/endpoint/sagas/loadCities.js
+++ b/src/modules/endpoint/sagas/loadCities.js
@@ -1,0 +1,33 @@
+// @flow
+
+import type { Saga } from 'redux-saga'
+import { createCitiesEndpoint, Payload } from '@integreat-app/integreat-api-client'
+import { call, put } from 'redux-saga/effects'
+import request from '../request'
+import { baseUrl } from '../constants'
+import type { DataContainer } from '../DataContainer'
+import CityModel from '@integreat-app/integreat-api-client/models/CityModel'
+import type { PushCitiesActionType } from '../../app/StoreActionType'
+
+function * loadCities (
+  dataContainer: DataContainer,
+  forceRefresh: boolean
+): Saga<Array<CityModel>> {
+  let cities: Array<CityModel>
+  if (!forceRefresh && (yield call(() => dataContainer.citiesAvailable()))) {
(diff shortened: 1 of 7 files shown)


src/modules/endpoint/sagas/loadCities.js
```diff
@@ -0,0 +1,33 @@
...
+
+function * loadCities (
+  dataContainer: DataContainer,
+  forceRefresh: boolean
+): Saga<Array<CityModel>> {
+  let cities: Array<CityModel>
+  if (!forceRefresh && (yield call(() => dataContainer.citiesAvailable()))) {
+    cities = yield call(() => dataContainer.getCities())
+  } else {
+    const payload: Payload<Array<CityModel>> = yield call(() => request(createCitiesEndpoint(baseUrl)))
```
Simple-codinger: Maybe we should check if there is a valid internet connection

Schedulaar: Then the request fails. If that happens, an error will appear.
Once NATIVE-334 is finished, the app will instead of showing an error, use old data to display meaningful content.

src/modules/endpoint/sagas/loadCities.js
```diff
@@ -0,0 +1,33 @@
...
+import type { DataContainer } from '../DataContainer'
+import CityModel from '@integreat-app/integreat-api-client/models/CityModel'
+import type { PushCitiesActionType } from '../../app/StoreActionType'
+
+function * loadCities (
+  dataContainer: DataContainer,
+  forceRefresh: boolean
+): Saga<Array<CityModel>> {
+  let cities: Array<CityModel>
+  if (!forceRefresh && (yield call(() => dataContainer.citiesAvailable()))) {
```
maxammann:
```suggestion
  const citiesAvailable = yield call(() => dataContainer.citiesAvailable())
  if (!forceRefresh && citiesAvailable) {
```

Schedulaar: Hmm... if forceRefresh is true this might cause an unnecessary fs call, therefore I'd like to leave it like this as a mini optimzation vs. a mini code readability improvement :sweat_smile:

PR #123, opened 2019-08-21
NATIVE-324 and NATIVE-327: Test database and first sagas

Simple-codinger:


src/modules/endpoint/__tests__/DatabaseConnector.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+    // store city
+    await dbCon.storeCities([testCity])
+
+    return dbCon.loadCities().then(cities => {
+      expect(cities).toStrictEqual([testCity])
+    })
+  })
+  it('should store/load lastUpdate', async () => {
+    const currentMoment = moment.tz('UTC')
```
Schedulaar: We usually use fixed timestamps (not the current one) to make the tests determinstic...

src/modules/endpoint/__tests__/DatabaseConnector.spec.js
```diff
@@ -0,0 +1,308 @@
...
+import DatabaseConnector from '../DatabaseConnector'
+import DatabaseContext from '../DatabaseContext'
+import RNFetchBlob from 'rn-fetch-blob'
+import moment from 'moment-timezone'
+
+jest.mock('rn-fetch-blob')
+const databaseConnector = new DatabaseConnector()
+
+beforeEach(() => {
+  // $FlowFixMe
```
klinzo: Add a note why a FlowFixMe is needed.

src/modules/endpoint/__tests__/DatabaseConnector.spec.js
```diff
@@ -0,0 +1,308 @@
...
+    sortingName: 'testCity',
+    prefix: 'Stadt'
+  })
+  const testCategory = new CategoryModel({
+    root: true,
+    path: 'test/path',
+    title: 'testCategory',
+    content: 'test content',
+    thumbnail: 'thumbnail.png',
+    parentPath: null,
```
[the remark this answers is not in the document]
maxammann: The detection of these errors will be fixed by: https://issues.integreat-app.de/browse/APICLIENT-22

src/modules/endpoint/__tests__/DatabaseConnector.spec.js
```diff
@@ -0,0 +1,308 @@
...
+    })
+  })
+  describe('loadCities', () => {
+    it('should throw exception if cities are not persisted', () => {
+      expect(databaseConnector.loadCities()).rejects.toThrowError()
+    })
+    it('should return a value that matches the one that was stored', async () => {
+      await databaseConnector.storeCities([testCity])
+
+      return databaseConnector.loadCities().then(cities => {
```
maxammann: you can use here async/await syntax:
```
    it('should return a value that matches the one that was stored', async () => {
      await databaseConnector.storeCities([testCity])

      const cities = await databaseConnector.loadCities()
      expect(cities).toStrictEqual([testCity])
    })
```

There are more cases in this test.

__mocks__/rn-fetch-blob.js
```diff
@@ -9,11 +52,15 @@ export default {
     }
   },
   fs: {
+    readdirSync: readdirSync,
     exists: existsMock,
+    writeFile: writeMockFile,
+    readFile: readMockFile,
```
maxammann:
```suggestion
    readFile:  jest.fn(readMockFile),
```

Add jest function mocks to all the exported functions here.

src/modules/endpoint/__tests__/DefaultDataContainer.spec.js
```diff
@@ -0,0 +1,42 @@
...
+      const defaultDataContainer = new DefaultDataContainer()
+      expect(defaultDataContainer.isCached('cities', new DatabaseContext())).toBe(false)
+    })
+  })
+  it('should look at the file system if data is not persisted in the cache', async () => {
+    const defaultDataContainer = new DefaultDataContainer()
+    await defaultDataContainer.setCities([testCity])
+
+    const anotherDataContainer = new DefaultDataContainer()
+    const readFile = jest.spyOn(RNFetchBlob.fs, 'readFile')
```
maxammann: If you add the jest function mocks (jest.fn()) in to the RNFetchBlob mock then you can just assert directly:
`expect(RNFetchBlob.fs.readFile).toHaveBeenCalled()`

src/modules/endpoint/__tests__/DatabaseConnector.spec.js
```diff
@@ -0,0 +1,308 @@
...
+      return databaseConnector.loadResourceCache(context).then(cache => {
+        expect(cache).toEqual({})
+      })
+    })
+    it('should return a value that matches the one that was stored', async () => {
+      const context = new DatabaseContext('tcc', 'de')
+      await databaseConnector.storeResourceCache(testResources, context)
+
+      return databaseConnector.loadResourceCache(context).then(cache => {
+        expect(cache).toMatchSnapshot()
```
maxammann: I think testing this without the snapshots like you did in the other tests is better. I think `testResources` should be the same as the `cache`

__mocks__/rn-fetch-blob.js
```diff
@@ -1,5 +1,48 @@
-const existsMock = jest.fn()
-existsMock.mockReturnValueOnce({ then: jest.fn() })
+const path = require('path')
+
+let mockFiles = Object.create(null)
```
Schedulaar:
```suggestion
let mockFiles = {}
```

__mocks__/rn-fetch-blob.js
```diff
@@ -1,5 +1,48 @@
...
+
+function deleteAllMockFiles () {
+  mockFiles = Object.create(null)
```
Schedulaar:
```suggestion
  mockFiles = {}
```

src/modules/endpoint/__tests__/DatabaseConnector.spec.js
Simple-codinger: If I do it without snapshot testing, then there is a problem with lastUpdate (lastUpdate is a moment object in the expected data, but a string in the received data)

maxammann: Right, that is actually a bug right now. It will be resolved by the mapping for resource cache: https://issues.integreat-app.de/browse/NATIVE-330

Go with this such that the snapshot test doesn't stay
```suggestion
expect(cache).toEqual({
        'de': {
          '/path/to/page': {
            'https://test.de/path/to/resource/test.png': {
              'filePath': '/local/path/to/resource/b4b5dca65e423.png',
              'hash': 'testHash',
              'lastUpdate': '2011-02-04T23:00:00.000Z'
            }
          },
          '/path/to/page/child': {
            'https://test.de/path/to/resource/test2.jpg': {
              'filePath': '/local/path/to/resource/970c65c41eac0.jpg',
              'hash': 'testHash',
              'lastUpdate': '2011-02-04T23:00:00.000Z'
            }
          }
        }
      })
```

maxammann: Maybe also add a comment

src/modules/endpoint/__tests__/DefaultDataContainer.spec.js
```diff
@@ -0,0 +1,41 @@
...
+  name: 'testCityName',
+  code: 'tcc',
+  live: true,
+  eventsEnabled: true,
+  extrasEnabled: true,
+  sortingName: 'testCity',
+  prefix: 'Stadt'
+})
+
+describe('DefaultDataContainer', () => {
```
maxammann: Tests for a lot of functions are missing here.

A test which is needed here, should check whether the loading for different cities/languages works.
This means one should be able to get the languages for a city A after getting them for city B.

src/modules/endpoint/__tests__/DefaultDataContainer.spec.js
```diff
@@ -17,6 +28,118 @@ const testCity = new CityModel({
   prefix: 'Stadt'
 })
 
+const testLanguage = new LanguageModel({
```
maxammann: I added yesterday a few builders for test data: https://github.com/Integreat/integreat-react-native-app/commit/e66f4fa0d1ed9be7793489030b734a55a7d13caa

Would be cool if you can use them here too :)

maxammann: If they are fine for you, you can just cherry-pick that commit

Simple-codinger: Looks good, I will use it

src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -73,7 +73,7 @@ export default function * loadCityContent (
 
     const fetchMap = resourceURLFinder.buildFetchMap(
       categoriesMap.toArray().concat(events),
-      (url, path) => buildResourceFilePath(url, path, newCity)
+      (url, path, urlHash) => buildResourceFilePath(url, path, newCity, urlHash)
```
Schedulaar: I don't get why you added the urlHash here. It's not used in the resourceUrlFinder.

Schedulaar:
Both this PR as well as https://github.com/Integreat/integreat-react-native-app/pull/133 are super messed up with commits of each other. Is there a way to clean this before merging? I'm quite lost in the diffs.

src/modules/endpoint/ResourceURLFinder.js
```diff
@@ -55,8 +56,8 @@ export default class ResourceURLFinder {
       return {
         ...acc,
         ...keyBy(
-          Array.from(urlSet).map(url => [url, path]),
-          ([url, path]) => buildFilePath(url, path)
+          Array.from(urlSet).map(url => [url, path, hashUrl(url)]),
```
Schedulaar: Why did you move the hashing to the ResourceURLFinder?

maxammann:
I think you can change the base to my 327 branch.

maxammann:
@Simple-codinger seems to work, right? not sure though whether all changes are included

src/modules/endpoint/ResourceURLFinder.js
maxammann: The hash URL was missing in the return type of fetchResources(). In order to fix this I had to generate it earlier.

src/modules/endpoint/__tests__/ResourceURLFinder.spec.js
```diff
@@ -44,7 +44,8 @@ describe('ResourceURLFinder', () => {
       },
       { path: '/path2', thumbnail: '', content: `<img src="https://ex.am/pl2.png" alt="Crazy" />` }
     ]
-    const fetchMap = finder.buildFetchMap(input, (url, path) => `buildFilePath('${url}', '${path}')`)
+    const fetchMap = finder.buildFetchMap(input, (url, path, urlHash) =>
+      `buildFilePath('${url}', '${path}', '${path}')`)
```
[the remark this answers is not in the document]
maxammann: Fixed on my branch

src/modules/endpoint/sagas/loadCityContent.js
maxammann: It is used in fetchResourceCache. This is visible in NATIVE-328

src/testing/ThrowingDataContainer.js
```diff
@@ -0,0 +1,74 @@
+// @flow
```
Schedulaar: I cannot see it used anywhere.

src/modules/endpoint/buildResourceFilePath.js
```diff
@@ -1,10 +1,8 @@
 // @flow
 
-import fnv from 'fnv-plus'
 import { getResourceCacheFilesDirPath } from '../platform/constants/webview'
 import getExtension from './getExtension'
 
-export default (url: string, path: string, city: string) => {
-  const urlHash = fnv.hash(url).hex()
+export default (url: string, path: string, city: string, urlHash: string) => {
   return `${getResourceCacheFilesDirPath(city)}/${urlHash}.${getExtension(url)}`
```
Schedulaar: This does not correspond to the documentation in the docs (see persistence).
Also I think we don't need to have a urlHash in the resourceCache, since the actual hash is irrelevant (and only the filepath is the necessary information).

src/modules/endpoint/sagas/__tests__/fetchResourceCache.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    }, resources)
+
+    delete lessStrictExpected['/augsburg/de/category_0-0'] /* The first category is excluded because an the
+                                                              FetcherModule produced an error for this */
+    expect(fetchedResources).toMatchObject(lessStrictExpected)
+    expect(spy).toHaveBeenCalledWith(expect.stringContaining('Failed to download https://integreat/title_0-0-300x300.png'))
+    spy.mockRestore()
+  })
+
+  it('should should yield error if fetching fails', () => {
```
Schedulaar:
```suggestion
  it('should yield error if fetching fails', () => {
```

src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -62,26 +62,6 @@ function * fetchCategory (dataContainer: DataContainer, action: FetchCategoryAct
...
-      type: 'FETCH_CATEGORY',
-      params: {
-        language: newLanguage,
-        path: `/${action.params.city}/${newLanguage}`,
-        ...action.params
-      }
-    }
-    yield put(newFetchCategory)
-  }
-}
```
Schedulaar: Why did you remove this?

maxammann:
Unfortunately again WIP.

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -82,7 +82,7 @@ class DefaultDataContainer implements DataContainer {
   }
 
   getResourceCache = async (city: string, language: string): Promise<LanguageResourceCacheStateType> => {
-    const context = new DatabaseContext(city, language)
+    const context = new DatabaseContext(city, null)
```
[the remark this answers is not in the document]
Simple-codinger: The resources should be available in all languages, that is why we have to set the language to null

src/modules/app/containers/__tests__/NavigatorContainer.spec.js
```diff
@@ -21,6 +21,7 @@ jest.mock('../../../layout/containers/HeaderContainer', () => {
 })
 jest.mock('@react-native-community/async-storage')
 jest.mock('react-native-share')
+jest.mock('rn-fetch-blob')
```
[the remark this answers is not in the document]
Simple-codinger: Something in this test uses RnFetchBlob.DocumentDir, that's why the test would fail if jest.mock('rn-fetch-blob') is not there

src/modules/endpoint/sagas/__tests__/fetchResourceCache.spec.js
klinzo: Isn't it rather `should put error` than `should yield error`?

src/modules/endpoint/sagas/__tests__/watchContentLanguageSwitch.spec.js
```diff
@@ -0,0 +1,101 @@
...
+  beforeEach(async () => {
+    RNFetchBlob.fs._reset()
+    await AsyncStorage.clear()
+  })
+
+  const city = 'augsburg'
+  const newLanguage = 'ar'
+
+  describe('fetchCategory', () => {
+    it('should yield an action when switching language', async () => {
```
klinzo:
```suggestion
    it('should put an action when switching language', async () => {
```

src/modules/endpoint/sagas/__tests__/watchFetchCategory.spec.js
```diff
@@ -0,0 +1,249 @@
...
+describe('watchFetchCategories', () => {
+  beforeEach(() => {
+    RNFetchBlob.fs._reset()
+  })
+
+  const city = 'augsburg'
+  const language = 'en'
+
+  describe('fetchCategory', () => {
+    it('should yield an action which pushes the categories', async () => {
```
klinzo:
```suggestion
    it('should put an action to push categories', async () => {
```

src/modules/endpoint/sagas/__tests__/watchFetchCategory.spec.js
```diff
@@ -0,0 +1,249 @@
...
+          cityContent: {
+            city: city
+          }
+        })
+        .not.put.like({ action: { type: 'PUSH_CATEGORY' } })
+        .not.put.like({ action: { type: 'FETCH_CATEGORY_FAILED' } })
+        .run()
+    })
+
+    it('should try to retrieved if peeking (independent of language)', async () => {
```
klinzo: Not quite sure what you want to say here

klinzo: I don't get what you are testing for here

src/modules/endpoint/sagas/__tests__/watchFetchCategory.spec.js
```diff
@@ -0,0 +1,249 @@
...
+          params: {
+            message: 'Error in fetchCategory: Jemand hat keine 4 Issues geschafft!',
+            key: 'categories-key'
+          }
+        })
+        .run()
+    })
+  })
+
+  it('should correctly call fetchEvent when triggered', async () => {
```
klinzo:
```suggestion
  it('should correctly call fetchCategory when triggered', async () => {
```

src/modules/endpoint/sagas/__tests__/watchFetchCities.spec.js
```diff
@@ -0,0 +1,68 @@
...
+jest.mock('rn-fetch-blob')
+jest.mock('../loadCities')
+
+describe('watchFetchCities', () => {
+  beforeEach(() => {
+    RNFetchBlob.fs._reset()
+  })
+
+  describe('fetchCities', () => {
+    it('should yield an action which pushes the cites', async () => {
```
klinzo:
```suggestion
    it('should yield an action which pushes the cities', async () => {
```

src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -40,15 +36,18 @@ export default function * fetchResourceCache (
     }
 
     const targetCategories: { [categoryPath: PathType]: Array<FilePathType> } =
-      invertBy(mapValues(fetchMap, ([url, path]) => path))
+      invertBy(mapValues(fetchMap, ([url, path, urlHash]) => path))
 
     const resourceCache = mapValues(targetCategories, filePaths =>
       reduce(filePaths, (acc, filePath) => {
         const downloadResult = successResults[filePath]
+        const [, , urlHash] = fetchMap[filePath]
```
klinzo: Maybe
```suggestion
        const urlHash = fetchMap[filePath][2]
```
would be clearer. But looks okay too

src/modules/endpoint/selectors/__tests__/isPeekingRoute.spec.js
```diff
@@ -0,0 +1,35 @@
+// @flow
+
+import isPeekingRoute from '../isPeekingRoute'
+
+describe('isPeekingRoute', () => {
+  it('should return true if new city is equal to content city', () => {
```
klinzo: You're expecting false, so either change the description or the test. Also a test for the opposite is missing.

src/modules/endpoint/__tests__/DefaultDataContainer.spec.js
```diff
@@ -0,0 +1,223 @@
...
+  it('should return an empty object if no resources where found', async () => {
+    const defaultDataContainer = new DefaultDataContainer()
+    await defaultDataContainer.setResourceCache('testCity', 'de', testResources)
+    const result = await defaultDataContainer.getResourceCache('testCity', 'en')
+    expect(result).toBeEmpty()
+  })
+  it('should return the lastUpdateMoment associated with the context', async () => {
+    const defaultDataContainer = new DefaultDataContainer()
+    const lastUpdate = moment('20110205')
+    const anotherLastUpdate = moment('20110305')
```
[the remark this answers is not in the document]
klinzo: @Simple-codinger you should use timezones everywhere

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,108 @@
...
+    if (depth === this._depth) {
+      return
+    }
+
+    for (let i = 0; i < this._maxWidth; i++) {
+      const path = `${category.path}/category_${depth}-${i}`
+      const lastUpdate = moment('2017-11-18T19:30:00.000Z', moment.ISO_8601)
+
+      const newChild = new CategoryModel({
+        id: 0,
```
klinzo: Maybe using different ids would make it easier to distinguish categories sometimes

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,108 @@
...
+        path,
+        title: `Category ${depth}-${i}`,
+        content: `<h1>This is a sample page</h1>
+                    <img src="https://integreat/title_${depth}-${i}-300x300.png"/>
+                    <p>This is a sample page</p>
+                    <img src="https://integreat/category_${depth}-${i}-300x300.png"/>`,
+        order: -1,
```
Schedulaar: you should use `i` here.

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,108 @@
...
+    return {
+      [url]: {
+        filePath: `path/to/documentDir/resource-cache/v1/some-city/files/${hash}.png`,
+        lastUpdate: moment(lastUpdate).add(this.predictableNumber(index), 'days'),
+        hash
+      }
+    }
+  }
+
+  addChildren (
```
Schedulaar: I guess this method should be "private" (and static?) as well?

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,108 @@
...
+import seedrandom from 'seedrandom'
+import hashUrl from '../../modules/endpoint/hashUrl'
+
+const DEFAULT_MAX_WIDTH = 3
+const DEFAULT_DEPTH = 2
+const MAX_PREDICTABLE_VALUE = 6
+
+class CategoriesMapModelBuilder {
+  _depth: number
+  _maxWidth: number
```
Schedulaar: it's the `arity` of the tree. You can add a comment that the builder generates a perfect tree of categories.

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,108 @@
...
+    category: CategoryModel,
+    categories: Array<CategoryModel>,
+    resourceCache: LanguageResourceCacheStateType,
+    depth: number) {
+    if (depth === this._depth) {
+      return
+    }
+
+    for (let i = 0; i < this._maxWidth; i++) {
+      const path = `${category.path}/category_${depth}-${i}`
```
Schedulaar: The `depth` here is quite redundant and should be removed.

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,108 @@
...
+        id: 0,
+        path,
+        title: `Category ${depth}-${i}`,
+        content: `<h1>This is a sample page</h1>
+                    <img src="https://integreat/title_${depth}-${i}-300x300.png"/>
+                    <p>This is a sample page</p>
+                    <img src="https://integreat/category_${depth}-${i}-300x300.png"/>`,
+        order: -1,
+        availableLanguages: new Map(),
+        thumbnail: `http://thumbnails/category_${depth}-${i}.png`,
```
Schedulaar: That way many categories get the same thumbnail assigned:
```
                         t_0-0
         /                 |                 \
      t_1-0              t_1-1              t_1-2
   /    |    \        /    |    \        /    |    \
t_2-0 t_2-1 t_2-2  t_2-0 t_2-1 t_2-2  t_2-0 t_2-1 t_2-2
```
(which might not be a huge problem)
Also for the  images above in the content
But the lastUpdate of these differ in the resourceCache which is odd I guess.

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,108 @@
...
+    const categories = []
+    const resourceCache = {}
+
+    for (let i = 0; i < this._maxWidth; i++) {
+      this.addChildren(new CategoryModel({
+        id: 0,
+        path: '/augsburg/de',
+        title: 'augsburg',
+        content: '',
+        order: -1,
```
Schedulaar:
```suggestion
        order: i,
```

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,108 @@
...
+        parentPath: '',
+        lastUpdate: moment('2017-11-18T19:30:00.000Z', moment.ISO_8601)
+      }), categories, resourceCache, 0)
+    }
+
+    return { categories: new CategoriesMapModel(categories), resourceCache }
+  }
+}
+
+export default CategoriesMapModelBuilder
```
Schedulaar: I'd actually suggest testing this, since it's quite complex. A single snapshot might do the trick.

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,108 @@
...
+
+  build (): CategoriesMapModel {
+    return this.buildAll().categories
+  }
+
+  buildAll (): { categories: CategoriesMapModel, resourceCache: { [path: string]: FileCacheStateType } } {
+    const categories = []
+    const resourceCache = {}
+
+    for (let i = 0; i < this._maxWidth; i++) {
```
Schedulaar: Is this `for` loop necessary? Have the feeling that it's not (and that it's actually too much)

src/testing/builder/EventModelBuilder.js
```diff
@@ -0,0 +1,95 @@
...
+    return Array.from({ length: this._eventCount }, (x, index) => {
+      const mockDate = moment('2015-01-01T00:00:00.000Z', moment.ISO_8601)
+      const startDate = moment(mockDate.add(this.predictableNumber(index), 'years').toISOString(), moment.ISO_8601)
+      const endDate = moment(mockDate.add(this.predictableNumber(index), 'hours').toISOString(), moment.ISO_8601)
+      const lastUpdate = moment(mockDate.subtract(this.predictableNumber(index), 'months').toISOString(), moment.ISO_8601)
+      const path = `/augsburg/en/events/event${index}`
+      return {
+        path,
+        event: new EventModel({
+          id: index + 1,
```
Schedulaar: why is the +1?

src/testing/builder/EventModelBuilder.js
```diff
@@ -0,0 +1,95 @@
...
+          lastUpdate,
+          content: `<h1>This is a sample event</h1>
+                    <img src="https://integreat/title_${index}-300x300.png"/>
+                    <p>This is a sample event</p>
+                    <img src="https://integreat/event_${index}-300x300.png"/>`,
+          thumbnail: 'https://integreat/thumbnail.png'
+        }),
+        resources: {
+          ...this.createResource(`https://integreat/title_${index}-300x300.png`, index, lastUpdate),
+          ...this.createResource(`https://integreat/event_${index}-300x300.png`, index, lastUpdate)
```
Schedulaar: the thumbnail is missing in the resources

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,108 @@
...
+        lastUpdate
+      })
+
+      resourceCache[path] = {
+        ...this.createResource(`https://integreat/title_${depth}-${i}-300x300.png`,
+          depth * this._maxWidth + i,
+          lastUpdate),
+        ...this.createResource(`https://integreat/category_${depth}-${i}-300x300.png`,
+          depth * this._maxWidth + i,
+          lastUpdate)
```
Schedulaar: the thumbnail is missing in the resources

src/testing/builder/LanguageModelBuilder.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+
+import { LanguageModel } from '@integreat-app/integreat-api-client'
+
+const languages = [
+  new LanguageModel('en', 'English'),
+  new LanguageModel('de', 'Deutsch')
```
Schedulaar: an RTL language with strange chars (like ar/fa) would be nice

src/testing/builder/CategoriesMapModelBuilder.js
maxammann: I think it is necessary to have have test data with more than one children. Ofc the trees get really huge, but this should be no problem in terms of computing power.

src/testing/builder/CategoriesMapModelBuilder.js
maxammann: You are right, I created a task: https://issues.integreat-app.de/browse/NATIVE-350

src/testing/builder/EventModelBuilder.js
maxammann: Events do not have the id 0 afaik. The id 0 is special in WP.

src/modules/fetcher/__mocks__/FetcherModule.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+    FetcherModule.currentlyFetching = true
+
+    for (let i = 0; i < 100; i++) {
+      progress(i / 100)
+    }
+
+    FetcherModule.currentlyFetching = false
+
+    let errorMessage = 'This result is invalid because it is the first result produced by the FetcherModule.js mock.'
```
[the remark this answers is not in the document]
maxammann: I switched to a pseudo random path and adjusted the test.

src/testing/builder/CategoriesMapModelBuilder.js
maxammann: I'll try to include the names of the ancestors in each thumbnail path

src/modules/endpoint/__tests__/DatabaseConnector.spec.js
```diff
@@ -0,0 +1,299 @@
...
+        {
+          filePath: '/local/path/to/resource/970c65c41eac0.jpg',
+          lastUpdate: moment('2011-02-04T23:00:00.000Z', moment.ISO_8601),
+          hash: 'testHash'
+        }
+      }
+    }
+  }
+
+  describe('getContentPath', () => {
```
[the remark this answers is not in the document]
Simple-codinger: This convention would be helpful

src/modules/endpoint/__tests__/DatabaseConnector.spec.js
```diff
@@ -0,0 +1,299 @@
...
+      await databaseConnector.storeEvents(testEvent, context)
+
+      const events = await databaseConnector.loadEvents(context)
+      expect(events).toEqual(testEvent)
+    })
+  })
+
+  describe('storeResourceCache', () => {
+    it('should throw error if data is empty', () => {
+      const context = new DatabaseContext('tcc', 'de')
```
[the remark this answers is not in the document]
Simple-codinger: Ok, found the problem, you should not use resolve in combination with toThrowError. I will delete this test, it doesn't make any sense, you should indeed be able to save an empty resource cache

src/modules/endpoint/__tests__/DefaultDataContainer.spec.js
```diff
@@ -0,0 +1,223 @@
...
+    it('should not delete any data if there are no previous resources available', async () => {
+      const defaultDataContainer = new DefaultDataContainer()
+      await defaultDataContainer.setResourceCache('testCity', 'de', testResources)
+
+      expect(RNFetchBlob.fs.unlink).not.toHaveBeenCalled()
+    })
+    it('should unlink the outdated resources if there are new resources available', async () => {
+      const defaultDataContainer = new DefaultDataContainer()
+      await defaultDataContainer.setResourceCache('testCity', 'en', anotherTestResources)
+      await defaultDataContainer.setResourceCache('testCity', 'de', previousResources)
```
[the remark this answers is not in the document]
Simple-codinger: you are right! The first call is unnecessary

src/testing/builder/EventModelBuilder.js
Schedulaar: In fact, EventModel as well as CategoryModel don't have any numeric ids :D

src/modules/endpoint/sagas/__tests__/loadLanguages.spec.js
```diff
@@ -8,19 +8,24 @@ import LanguageModelBuilder from '../../../../testing/builder/LanguageModelBuild
...
-      .build()
-  }
-)
+jest.mock('@integreat-app/integreat-api-client',
+  () => {
+    const actual = jest.requireActual('@integreat-app/integreat-api-client')
+    return {
+      ...actual,
+      createLanguagesEndpoint: () => {
+        const { EndpointBuilder } = require('@integreat-app/integreat-api-client')
```
Schedulaar: Well, that's funny :D you require a module that you're about to mock and you're actually requiring the mock.
You should probably use `const EndpointBuilder = actual.EndpointBuilder`.

src/modules/fetcher/__mocks__/FetcherModule.js
```diff
@@ -25,13 +25,18 @@ class FetcherModule {
...
+      value => ({ lastUpdate: moment('2016-02-01T10:35:20Z'), url: value, errorMessage: null })
+    )
 
-    return Promise.resolve(reduce(targetFilePaths, (result, value, key) => {
-      result[key] = { lastUpdate: moment('2016-02-01T10:35:20Z'), url: value, errorMessage: errorMessage }
-      errorMessage = null
-      return result
-    }, {}))
+    const paths = Object.keys(fetchResult)
+    const pseudoRandomPath = paths[Math.floor(0.7 * paths.length)]
```
Schedulaar: the problem here is, that the order of `Object.keys` is arbitrary and therefore your "pseudo" random operation is also arbitrary and can be regarded as actually random.

src/modules/endpoint/sagas/__tests__/loadEvents.spec.js
```diff
@@ -8,20 +8,25 @@ import EventModelBuilder from '../../../../testing/builder/EventModelBuilder'
...
-      .build()
-  }
-)
+jest.mock('@integreat-app/integreat-api-client',
+  () => {
+    const actual = jest.requireActual('@integreat-app/integreat-api-client')
+    return {
+      ...actual,
+      createEventsEndpoint: () => {
+        const { EndpointBuilder } = require('@integreat-app/integreat-api-client')
```
Schedulaar: Here you can probably use `const EndpointBuilder = actual.EndpointBuilder`.

src/modules/endpoint/sagas/__tests__/loadCities.spec.js
```diff
@@ -8,19 +8,24 @@ import CityModelBuilder from '../../../../testing/builder/CitiyModelBuilder'
...
-      .build()
-  }
-)
+jest.mock('@integreat-app/integreat-api-client',
+  () => {
+    const actual = jest.requireActual('@integreat-app/integreat-api-client')
+    return {
+      ...actual,
+      createCitiesEndpoint: () => {
+        const { EndpointBuilder } = require('@integreat-app/integreat-api-client')
```
Schedulaar: Here you can probably use `const EndpointBuilder = actual.EndpointBuilder`.

src/modules/endpoint/sagas/__tests__/loadCategories.spec.js
```diff
@@ -8,19 +8,24 @@ import CategoriesMapModelBuilder from '../../../../testing/builder/CategoriesMap
...
-      .build()
-  }
-)
+jest.mock('@integreat-app/integreat-api-client',
+  () => {
+    const actual = jest.requireActual('@integreat-app/integreat-api-client')
+    return {
+      ...actual,
+      createCategoriesEndpoint: () => {
+        const { EndpointBuilder } = require('@integreat-app/integreat-api-client')
```
Schedulaar: Here you can probably use `const EndpointBuilder = actual.EndpointBuilder`.

src/modules/endpoint/sagas/__tests__/fetchResourceCache.spec.js
```diff
@@ -44,21 +43,33 @@ describe('fetchResourceCache', () => {
...
+    }
+
+    const expectedCount = {
+      ...resources['/augsburg/de/category_0'],
+      ...resources['/augsburg/de/category_0/category_0']
+    }
+
+    expect(Object.keys(fetchedCount)).toHaveLength(
+      Object.keys(expectedCount).length - 1 /* The first url is excluded because the
+                                               FetcherModule mock produced an error for it */
```
Schedulaar: You cannot *really* say that it's the first url, since Object.keys' order is arbitrary. But you can say, that there's one entry that the FetcherModule produced an error for.

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -87,9 +95,8 @@ class CategoriesMapModelBuilder {
     const categories = []
     const resourceCache = {}
 
-    for (let i = 0; i < this._maxWidth; i++) {
-      this.addChildren(new CategoryModel({
-        id: 0,
+    for (let i = 0; i < this._arity; i++) {
```
Schedulaar: This `for` loop is wrong (what you can see as `i` is never used here).
You should call `this._addChildren(new CatModel(...))` (as you should also only create the root-model once)

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -28,50 +31,55 @@ class CategoriesMapModelBuilder {
...
     }
 
-    for (let i = 0; i < this._maxWidth; i++) {
-      const path = `${category.path}/category_${depth}-${i}`
+    for (let i = 0; i < this._arity; i++) {
+      const path = `${category.path}/category_${i}`
       const lastUpdate = moment('2017-11-18T19:30:00.000Z', moment.ISO_8601)
+      const id = depth * this._arity + i + 1
+      const resourceUrl1 = `https://integreat/title_${depth}-${i}-300x300.png`
+      const resourceUrl2 = `https://integreat/category_${depth}-${i}-300x300.png`
```
Schedulaar: The resourceUrls are still not unique. (Don't know if that's the expected behavior)

src/modules/endpoint/__tests__/DatabaseConnector.spec.js
```diff
@@ -0,0 +1,225 @@
...
+import EventModelBuilder from '../../../testing/builder/EventModelBuilder'
+
+jest.mock('rn-fetch-blob')
+const databaseConnector = new DatabaseConnector()
+
+beforeEach(() => {
+  RNFetchBlob.fs._reset()
+})
+
+describe('DatabaseConnector', () => {
```
Schedulaar: I still think testing where the `storeXyz` functions write their JSON files would be pretty helpful.

src/modules/endpoint/ResourceURLFinder.js
```diff
@@ -53,13 +58,18 @@ export default class ResourceURLFinder {
...
-        ...keyBy(
-          Array.from(urlSet).map(url => [url, path, hashUrl(url)]),
-          ([url, path, urlHash]) => buildFilePath(url, path, urlHash)
-        )
-      }
+      Array.from(urlSet).forEach(url => {
+        const urlHash = hashUrl(url)
+        const filePath = buildFilePath(url, path, urlHash)
+
+        if (!filePath) {
```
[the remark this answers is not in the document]
maxammann: I'm not only checking for an empty string here, I'm also checking for null/undefined.
As a developer you should have the possibility to to "filter" the urlSet by returning null in `buildFilePath`.

As this is currently not used I'll remove it

src/modules/endpoint/sagas/__tests__/fetchResourceCache.spec.js
```diff
@@ -44,21 +43,33 @@ describe('fetchResourceCache', () => {
 
     const fetchedResources = await dataContainer.getResourceCache(city, language)
 
-    const lessStrictExpected = transform(resources, (result, value, path) => {
-      forEach(value, (value, url) => {
-        result[path][url].lastUpdate = expect.any(Moment)
-      })
-      return result
-    }, resources)
+    if (fetchedResources.errorMessage !== undefined) {
```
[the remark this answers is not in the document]
maxammann: Flow reasons :man_shrugging:

src/modules/fetcher/__mocks__/FetcherModule.js
```diff
@@ -25,13 +25,18 @@ class FetcherModule {
 
     FetcherModule.currentlyFetching = false
 
-    let errorMessage = 'This result is invalid because it is the first result produced by the FetcherModule.js mock.'
+    const fetchResult = mapValues(
+      targetFilePaths,
+      value => ({ lastUpdate: moment('2016-02-01T10:35:20Z'), url: value, errorMessage: null })
```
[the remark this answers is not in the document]
maxammann: This is not a problem here as Zulu time is included in the ISO8601 string.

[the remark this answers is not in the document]
maxammann: This is not a problem here as Zulu time is included in the ISO8601 string.

src/modules/endpoint/sagas/__tests__/loadLanguages.spec.js
maxammann: This is not a problem as createLanguagesEndpoint is only called after the mock has been constructed :)

Else you would get a stack overflow :P

I think it is better to do it that way so that require() is called when the test code is run.

src/modules/fetcher/__mocks__/FetcherModule.js
maxammann: You are right. I sort now by the url. As time and errorMessage is the same for all values this should make it deterministic.

src/modules/endpoint/sagas/__tests__/fetchResourceCache.spec.js
maxammann: Forgot to adjust this. Thanks!

src/testing/builder/CategoriesMapModelBuilder.js
maxammann: This is not yet done.

maxammann: I switched to an other approach for generating unique resource urls

src/testing/builder/CategoriesMapModelBuilder.js
maxammann: True :man_facepalming:

maxammann:
I'm done again with fixing the comments regarding my changes.

src/modules/endpoint/__tests__/DatabaseConnector.spec.js
```diff
@@ -0,0 +1,225 @@
...
+      const context = new DatabaseContext('tcc', 'de')
+
+      const cache = await databaseConnector.loadResourceCache(context)
+      expect(cache).toEqual({})
+    })
+    // cache is not equal to testResources
+    // (lastUpdate is a moment object in the expected data, but a string in the received data)
+    // will be fixed in NATIVE-330
+    // eslint-disable-next-line jest/no-disabled-tests
+    it.skip('should return a value that matches the one that was stored', async () => {
```
maxammann: I added a link in the issue: https://issues.integreat-app.de/browse/NATIVE-330

src/modules/endpoint/DefaultDataContainer.js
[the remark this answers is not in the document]
maxammann: Would it be feasible to store only `LanguageResourceCacheStateType`s instead of the whole `CityResourceCacheStateType`?

That way we could use the Cache and add the language to the Context.

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -95,18 +96,17 @@ class CategoriesMapModelBuilder {
...
+      title: 'augsburg',
+      content: '',
+      order: -1,
+      availableLanguages: new Map(),
+      thumbnail: 'no_thumbnail',
+      parentPath: '',
+      lastUpdate: moment('2017-11-18T19:30:00.000Z', moment.ISO_8601)
+    }), resourceCache, 0)
 
     return { categories: new CategoriesMapModel(categories), resourceCache }
```
Schedulaar: As far as I can see, `categories` is always an empty array. You probably mean to use `this._categories` and remove the `categories` variable?

src/testing/builder/CategoriesMapModelBuilder.js
Schedulaar: Looking good :+1:

src/modules/endpoint/DefaultDataContainer.js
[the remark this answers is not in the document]
Schedulaar: Hmm... It's necessary though if we want to have a shared resource cache, since we need to check the resourceCaches of all other languages as well if we want to cleanup some resource.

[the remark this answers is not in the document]
Schedulaar: So... feasible yes, but not within this task :sweat_smile: We would need some other strategy for checking for resource-dependency (then maybe globally for all cities)

src/modules/endpoint/__tests__/DatabaseConnector.spec.js
Simple-codinger: I've now added some tests for the store functions

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,115 @@
...
+      const id = this._id
+      this._id++
+
+      const path = `${category.path}/category_${i}`
+      const lastUpdate = moment('2017-11-18T19:30:00.000Z', moment.ISO_8601)
+      const resourceUrl1 = `https://integreat/title_${id}-300x300.png`
+      const resourceUrl2 = `https://integreat/category_${id}-300x300.png`
+      const thumbnail = `http://thumbnails/category_${id}.png`
+
+      const newChild = new CategoryModel({
```
Schedulaar: you should define `hash` and `root` here as well.

src/testing/builder/CategoriesMapModelBuilder.js
```diff
@@ -0,0 +1,115 @@
...
+    return this.buildAll().categories
+  }
+
+  buildAll (): { categories: CategoriesMapModel, resourceCache: { [path: string]: FileCacheStateType } } {
+    const categories = []
+    const resourceCache = {}
+
+    this._categories = []
+    this._addChildren(new CategoryModel({
+      path: '/augsburg/de',
```
Schedulaar: you should define `hash` and `root` here as well.

src/testing/builder/EventModelBuilder.js
Schedulaar: Remove numeric id and add `hash`

src/modules/endpoint/buildResourceFilePath.js
Schedulaar: Please adjust the docs.

src/modules/endpoint/sagas/watchFetchCategory.js
Schedulaar: Changing language very fast is indeed possible on my emulator :thinking: I think this was the very reason why this code was introduced.

src/modules/endpoint/__tests__/buildResourceFilePath.spec.js
```diff
@@ -0,0 +1,32 @@
...
+describe('buildResourceFilePath', () => {
+  it('should return the resource file path', () => {
+    const urlString = 'https://ex.am/p.l/thumbnail.png'
+    const city = 'augsburg'
+    const hash = hashUrl(urlString)
+    expect(buildResourceFilePath(urlString, '/augsburg/de', city, hash))
+      .toBe('path/to/documentDir/resource-cache/v1/augsburg/files/81a74f17bb169f4dad2f59bb2e4670f9.png')
+  })
+
+  it('should ignore invalid extension', () => {
```
klinzo: No extension isn't really an invalid extension

src/modules/endpoint/sagas/__tests__/watchFetchCategory.spec.js
```diff
@@ -0,0 +1,249 @@
...
+            depth: 2,
+            key: 'categories-key',
+            language,
+            city
+          }
+        })
+        .run()
+    })
+
+    it('should do nothing if content can not be retrieved (e.g. language invalid) and not peeking', async () => {
```
klinzo:
```suggestion
    it('should do nothing if content can not be retrieved (e.g. language invalid and not peeking)', async () => {
```

src/modules/endpoint/sagas/__tests__/watchFetchCategory.spec.js
```diff
@@ -0,0 +1,249 @@
...
+        .withState({
+          cityContent: {
+            city: anotherCity
+          }
+        })
+        .provide({
+          call: (effect, next) => {
+            if (effect.fn === loadCityContent) {
+              if (!effect.args.includes(invalidLanguage)) {
+                throw new Error(`${invalidLanguage} wasn't fetched`)
```
klinzo: This check/throw isn't really necessary, right?

src/modules/endpoint/sagas/__tests__/watchFetchCategory.spec.js
```diff
@@ -0,0 +1,249 @@
...
+          cityContent: {
+            city: city
+          }
+        })
+        .not.put.like({ action: { type: 'PUSH_CATEGORY' } })
+        .not.put.like({ action: { type: 'FETCH_CATEGORY_FAILED' } })
+        .run()
+    })
+
+    it('should try to loadCityContent with an invalid language when peeking', async () => {
```
klinzo: Right now, this is kind of the same test as the one below (should put an error action) with a different error message. Wouldn't it be enough to just assert `loadCityContent` was called (http://redux-saga-test-plan.jeremyfairbank.com/integration-testing/partial-matching.html) if you just want to test that it also gets called if the language is invalid while peeking?

src/testing/builder/CategoriesMapModelBuilder.js
maxammann: This definitely needs tests :D

src/modules/endpoint/sagas/watchFetchCategory.js
klinzo: Did you check if this is still a problem?

PR #129, opened 2019-09-01
NATIVE-343: Test navigator

Schedulaar:


src/routes/landing/containers/LandingContainer.js
```diff
@@ -67,7 +67,7 @@ class LandingContainer extends React.Component<ContainerPropsType> {
     })
 
     navigation.navigate({
-      routeName: 'App',
+      routeName: 'CityContent',
```
maxammann: :+1: Did you test the navigation after the refactor (Modals, Nagivation back etc.)?

src/modules/app/containers/__tests__/NavigatorContainer.spec.js
```diff
@@ -0,0 +1,97 @@
...
+  beforeEach(() => {
+    AsyncStorage.clear()
+    jest.resetModules()
+  })
+
+  it('should pass fetchCategory to Navigator', () => {
+    jest.doMock('../../components/Navigator', () => () => null)
+    const store = mockStore({})
+    const result = TestRenderer.create(<Provider store={store}><NavigatorContainer /></Provider>)
+    const navigator = result.root.findByType(Navigator)
```
maxammann: ~This could be problematic here as the `Navigator` variable is not identical to the reference to the Navigator class used by the TestRenderer.~

src/modules/app/containers/__tests__/NavigatorContainer.spec.js
```diff
@@ -0,0 +1,97 @@
...
+const mockStore = configureMockStore()
+
+describe('NavigatorContainer', () => {
+  beforeEach(() => {
+    AsyncStorage.clear()
+    jest.resetModules()
+  })
+
+  it('should pass fetchCategory to Navigator', () => {
+    jest.doMock('../../components/Navigator', () => () => null)
```
maxammann: I think this mock is never actually used. The `Navigator` corresponds to the instance which was imported at the top of the file. Also because `NavigatorContainer` and therefore the Navigator  is imported before the doMock there is no way that mock can be used.

The solution is to import the NavigatorContainer and Navigator using `require`.

maxammann:
Hm I deleted the base branch of this PR...
Can you change the base?

src/modules/app/containers/__tests__/NavigatorContainer.spec.js
Schedulaar: Yup, thanks for discovering! Fixed it.

src/routes/landing/containers/LandingContainer.js
Schedulaar: Yup, tested it.

jest.config.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
     '<rootDir>/jest.setup.afterenv.js'
   ],
   'transformIgnorePatterns': [
-    'node_modules/(?!react-native|@react-native-community/async-storage|rn-fetch-blob|@integreat-app/integreat-api-client|antd|rc-.+)'
+    'node_modules/(?!react-native|@react-navigation/native/dist|react-navigation-stack|@react-native-community/async-storage|react-navigation-header-buttons|rn-fetch-blob|@integreat-app/integreat-api-client|antd|rc-.+)'
```
maxammann: Why are these ignores needed? ES versions shouldn't be the reason, right? Never touched this path afaik.
If applicable add some comments here.

Schedulaar: Yes, ES version is the reason. In @react-navigation there are some imports that require files under @react-navigation/native/dist, which are not transcribed correctly. Same for react-navigation-stack and react-navigation-header-buttons.

src/modules/app/components/__tests__/Navigator.spec.js
```diff
@@ -0,0 +1,80 @@
...
+        language: 'de',
+        key: expect.any(String)
+      })
+      expect(getByText('AppContainerOf(CityContent)')).toBeTruthy()
+      const key = mockCreateAppNavigationContainer.mock.calls[0][0].key
+      expect(fetchCategory).toHaveBeenCalledWith('augsburg', 'de', key)
+    })
+  })
+
+  it('should display Landing and call fetchCategory if no city is selected in settings', async () => {
```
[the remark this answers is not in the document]
Schedulaar: Nope, it wasn't. Thanks for spotting!

PR #130, opened 2019-09-02
NATIVE-117: Test error module

maxammann:


maxammann:
open for review

jest.setup.js
```diff
@@ -21,4 +15,24 @@ if (typeof window !== 'object') {
...
+
+function walkDir (dir, callback) {
+  fs.readdirSync(dir).forEach(f => {
+    const dirPath = path.join(dir, f)
+    const isDirectory = fs.statSync(dirPath).isDirectory()
+    isDirectory ? walkDir(dirPath, callback) : callback(path.join(dir, f))
+  })
+}
+
+// The following code automatically mocks the modules in `mocksPath`. This is required because jest mocks all these
```
Simple-codinger: shouldn't it be unmock?

src/modules/error/__tests__/ContentNotFoundError.js
```diff
@@ -0,0 +1,24 @@
...
+    expect(error.message).toMatchSnapshot()
+  })
+
+  it('should produce a stacktrace', () => {
+    try {
+      throw new ContentNotFoundError(
+        { type: 'category', id: '/augsburg/de/test', city: 'augsburg', language: 'language' }
+      )
+    } catch (error) {
+      expect(error.stack).toContain('ContentNotFoundError.js:16') /*  This expect depends on the order of the
```
Simple-codinger: Can we remove the line number here?

maxammann: I included it to assert that there is a stack trace. Not sure about an other way.

Schedulaar: You could use a regexp here and replace the line number with an arbitrary number.
Also the spec file should have a .spec.js ending

src/modules/error/containers/__tests__/FailureContainer.spec.js
```diff
@@ -0,0 +1,38 @@
...
+    const rendered = TestRenderer.create(<FailureContainerMock />)
+    const instance = rendered.root.findByType(FailureMock)
+    expect(instance.props).toEqual({
+      t: expect.anything(),
+      theme: expect.anything()
+    })
+  })
+
+  it('should render inner component correctly without passing props explicitly', () => {
+    jest.dontMock('../../components/Failure')
```
Simple-codinger: not really sure if you need to call dontMock

maxammann: Yes it is required. If it is missing then the mock of the other test is automatically applied.

src/modules/error/components/__tests__/Failure.spec.js
```diff
@@ -0,0 +1,43 @@
...
+import { fireEvent, render } from '@testing-library/react-native'
+import * as React from 'react'
+import { brightTheme } from '../../../theme/constants/theme'
+
+describe('Failure', () => {
+  it('should render a retry button if tryAgain is passed', () => {
+    const { getByTestId } = render(<Failure theme={brightTheme} error={new Error()} tryAgain={() => {}}
+                                            t={key => key} />)
+
+    expect(getByTestId('button-tryAgain')).not.toBeNull()
```
Schedulaar: wouldn't `expect(xyz).toBeTruthy()` be a safer way? Not sure wether getByTestId returns null or undefined or false if it's unsuccessfull... Also you seem to use .toBeTruthy in getByText below :thinking:

src/modules/error/containers/__tests__/FailureContainer.spec.js
```diff
@@ -0,0 +1,38 @@
...
+import * as React from 'react'
+import TestRenderer from 'react-test-renderer'
+import typeof FailureContainer from '../FailureContainer'
+
+jest.mock('react-i18next')
+jest.mock('../../../theme/hocs/withTheme')
+
+describe('FailureContainer', () => {
+  beforeEach(() => {
+    jest.resetModules()
```
Schedulaar: What is the resetModules for?

src/modules/platform/constants/locale.js
```diff
@@ -17,5 +17,3 @@ const getLocale = () => {
 
   return locale
 }
-
-export default getLocale
```
Schedulaar: Hmm... I don't really see that this is a constant...
Also: Why did you rename the file?

src/testing/customQueries.js
```diff
@@ -0,0 +1,18 @@
...
+import { isArray } from 'lodash'
+import { type NativeTestInstance } from '@testing-library/react-native'
+
+export function queryAllFlex (container: NativeTestInstance): Array<ReactTestInstance> {
+  return Array.from(container.findAll(c => c.getProp('style')))
+    .filter((node: NativeTestInstance) => {
+      const style = node.getProp('style')
+
+      if (!style) {
+        return
```
Schedulaar:
```suggestion
        return false
```

src/testing/utils.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+export const toHaveDirection = (received: Array<ReactTestInstance>) => {
+  if (received.every(instance => instance.props.style.some(style => style.flexDirection === 'row'))) {
+    return {
+      message: () =>
+        `expected received components not to have direction`,
```
Schedulaar: I guess the message shouldn't say the opposite if it passes as if it didn't
```suggestion
        `expected received components to have direction`,
```

src/testing/utils.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+export const toHaveDirection = (received: Array<ReactTestInstance>) => {
```
Schedulaar: not sure what this actually does... does it expect the array of ReactTestInstances to all have flex direction `row`? Then the naming should be adjusted.

docs/07-testing.md
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+```
+
+```js
+// FailureContainer.js
+export default withTheme()(translate('error')(Failure))
+```
+
+The most important part is to test the specific units using `@testing-library/react-native` and `react-test-renderer`. The following test is for the `Failure` component.
```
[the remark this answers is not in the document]
maxammann: yeah, the query utilities are better in @testing-library

jest.setup.afterenv.js
```diff
@@ -0,0 +1,3 @@
+import { toHaveDirection } from './src/testing/utils'
+
+expect.extend({ toHaveDirection })
```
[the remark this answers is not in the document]
maxammann: Flow does not support it: https://github.com/Integreat/integreat-react-native-app/pull/130/files#diff-3b9c618d74ecb99ace76a23c7f88e27eR18

Maybe we find a solution in the future...

src/modules/error/containers/__tests__/FailureContainer.spec.js
maxammann: It resets the mocked modules. If you now call jest.unmock() and import a module later it won't be mocked.

src/modules/platform/constants/locale.js
maxammann: Yeah probably, I just reverted the changes from @klinzo as there is an other way to inject the locale.

src/testing/utils.js
maxammann: I'm quite sure its that way around. Although this is very unintuitive.

src/testing/utils.js
maxammann: The naming is more generalized. It should check whether the supplied component(s) do have a direction in some way.

maxammann: I'll add a comment

src/testing/utils.js
maxammann: It is correct as it is. Tested it :)

src/modules/platform/constants/locale.js
maxammann: I rename this to getLocale.js

docs/07-testing.md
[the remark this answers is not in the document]
maxammann: Any objections?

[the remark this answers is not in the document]
Schedulaar: No

src/testing/utils.js
```diff
@@ -1,5 +1,12 @@
...
 
+/**
+ * Checks whether the resulting components all have direction information. This allows to test whether the component
+ * is internationalizable.
+ *
+ * @param received The received result
+ * @returns {{pass: boolean, message: (function(): string)}} A jest.extend compatible matcher
+ */
 export const toHaveDirection = (received: Array<ReactTestInstance>) => {
   if (received.every(instance => instance.props.style.some(style => style.flexDirection === 'row'))) {
```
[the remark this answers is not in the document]
maxammann: Yes the test were actually pretty stupid. I added now tests which test properly: https://github.com/Integreat/integreat-react-native-app/pull/130/files/b63bd9fdcab6d8bedcd1fcd3b82dfbf7338b3799..0670a68f590ed704b09ac43e77c9da55d4733bc0

I think testing for the RTL stuff is important as it is a key feature of the app (rtl support :))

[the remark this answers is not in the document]
maxammann: Just not sure whether that is the right way to go.

src/modules/i18n/containers/__tests__/I18nProviderContainer.spec.js
```diff
@@ -2,40 +2,34 @@
...
-      type: 'SET_CONTENT_LANGUAGE',
-      params: {
-        contentLanguage: 'de'
-      }
+    await waitForExpect(() => {
+      expect(store.getActions()).toEqual([{
+        type: 'SET_CONTENT_LANGUAGE',
+        params: {
+          contentLanguage: 'en'
+        }
```
Schedulaar: You don't expect that the right props are passed (which is what the title of the test says)

maxammann: I added the unit test

src/modules/i18n/containers/__tests__/I18nProviderContainer.spec.js
```diff
@@ -2,18 +2,41 @@
...
+    jest.doMock('../../components/I18nProvider', () => I18nProviderMock)
+    const I18nProviderContainerMock: I18nProviderContainer = require('../I18nProviderContainer').default
+
+    const store = mockStore({})
+    const rendered = TestRenderer.create(<Provider store={store}>
+      <I18nProviderContainerMock />
+    </Provider>)
+    const instance = rendered.root.findByType(I18nProviderMock)
+    expect(instance.props).toEqual({
+      setContentLanguage: expect.anything()
```
Schedulaar: one should test, that setContentLanguage dispatches an action, since this is part of the container logic
(call it, then check store.getActions)

src/modules/i18n/components/__tests__/I18nProvider.spec.js
```diff
@@ -42,29 +42,23 @@ describe('I18nProvider', () => {
...
     await waitForExpect(async () => {
       expect(mockSetContentLanguage).toHaveBeenCalledTimes(1)
-      expect(mockSetContentLanguage).toHaveBeenCalledWith('de')
-      expect(await appSettings.loadContentLanguage()).toBe('de')
+      expect(mockSetContentLanguage).toHaveBeenCalledWith('en')
+      expect(await appSettings.loadContentLanguage()).toBe('en')
     })
   })
 
   it('should not set content language if already set', () => {
```
Schedulaar: You have to use the `async () => {` syntax here and `await waitForExpect(async () => {`! Otherwise the test will always pass, even if the expects inside of `waitForExpect` don't!

src/modules/i18n/containers/__tests__/I18nProviderContainer.spec.js
maxammann: It is tested indirectly by the test below. but I added an explicit test

src/modules/i18n/components/__tests__/I18nProvider.spec.js
maxammann: True :P https://github.com/Integreat/integreat-react-native-app/commit/74566bdc002f8668b4aaa5521b13ed2a9bf1f990

PR #135, opened 2019-09-16
NATIVE-347: Switch to jenkins node

maxammann:


maxammann:
I'll just merge this as it does not need a review. 
**We are building now an an external VM. Please check the build times!**

PR #137, opened 2019-09-20
NATIVE-120: Test module fetcher

Simple-codinger:


src/modules/fetcher/__mocks__/NativeFetcherModule.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+
+import type { FetchResultType } from '../FetcherModule'
+import { NativeEventEmitter } from 'react-native'
+
+export type NativeFetcherModuleType = {
+  +addListener: (eventType: string) => void,
+  +removeListeners: (count: number) => void,
+  fetchAsync: (targetFilePaths: { [url: string]: string }) => Promise<FetchResultType>
```
Schedulaar: why did you make `fetchAsync` not read-only?

src/modules/fetcher/__tests__/FetcherModule.spec.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+
+import FetcherModule from '../FetcherModule'
+import NativeFetcherModule from '../NativeFetcherModule'
+import type { TargetFilePathsType } from '../FetcherModule'
+
+jest.mock('../NativeFetcherModule')
+
+beforeEach(() => {
```
Schedulaar: I guess it's better to expect a clean environment and clean it afterwards.
```suggestion
afterEach(() => {
```

src/modules/fetcher/__tests__/FetcherModule.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    await fetcherModule.fetchAsync(testTargetFilePaths, progress => console.log(progress))
+    expect(NativeFetcherModule.fetchAsync).toHaveBeenCalledTimes(1)
+    expect(NativeFetcherModule.fetchAsync).toHaveBeenCalledWith(testTargetFilePaths)
+  })
+  it('should not call fetchAsync if targetFiles are empty', async () => {
+    const fetcherModule = new FetcherModule()
+    await fetcherModule.fetchAsync({}, progress => console.log(progress))
+    expect(NativeFetcherModule.fetchAsync).not.toHaveBeenCalled()
+  })
+  it('should return an error if fetcher is already busy', async () => {
```
Schedulaar: Please test this by calling fetchAsync twice and have the promise of NativeFetcherModule.fetchAsync not resolve before the second call to fetchAsync.

Setting FetcherModule.currentlyFetching is an implementation detail

src/modules/fetcher/__tests__/FetcherModule.spec.js
```diff
@@ -0,0 +1,42 @@
...
+describe('FetcherModule', () => {
+  const testTargetFilePaths: TargetFilePathsType = {
+    'local/path/to/resource.png': 'http://randomtesturl.de/resource.png',
+    'local/path/to/resource2.jpg': 'http://randomtesturl.de/resource2.jpg'
+  }
+
+  it('currentlyFetching should be false if no fetching process is running', () => {
+    expect(FetcherModule.currentlyFetching).toBe(false)
+  })
+  it('should call fetchAsync with targetFiles on native module', async () => {
```
Schedulaar: Please also test the result of `fetchAsync`.

src/modules/fetcher/__tests__/FetcherModule.spec.js
```diff
@@ -0,0 +1,42 @@
...
+  jest.resetAllMocks()
+})
+
+describe('FetcherModule', () => {
+  const testTargetFilePaths: TargetFilePathsType = {
+    'local/path/to/resource.png': 'http://randomtesturl.de/resource.png',
+    'local/path/to/resource2.jpg': 'http://randomtesturl.de/resource2.jpg'
+  }
+
+  it('currentlyFetching should be false if no fetching process is running', () => {
```
Schedulaar: `currentlyFetching` is an implementation detail to me and should not be tested.

src/modules/fetcher/__mocks__/NativeFetcherModule.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+
+import type { FetchResultType } from '../FetcherModule'
+import { NativeEventEmitter } from 'react-native'
+
+export type NativeFetcherModuleType = {
```
maxammann: I think this type is not necessary.  If it is then use it directly in this file in line 12:

`const NativeFetcherModule:  NativeFetcherModuleType = {`

src/modules/fetcher/__tests__/FetcherModule.spec.js
maxammann: In case of resetAllMocks() I'd actually prefere to run it before each test. That way you can be sure that the function mocks are all reset for the very first test.

Note that resetAllMocks only affects `jest.fn` mocks

src/modules/fetcher/__tests__/FetcherModule.spec.js
maxammann: Only the effects of fetching twice should be tested.

You can do this by fetching twice and not resolving for one call.

src/modules/fetcher/__tests__/FetcherModule.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    expect(FetcherModule.currentlyFetching).toBe(false)
+  })
+  it('should call fetchAsync with targetFiles on native module', async () => {
+    const fetcherModule = new FetcherModule()
+    await fetcherModule.fetchAsync(testTargetFilePaths, progress => console.log(progress))
+    expect(NativeFetcherModule.fetchAsync).toHaveBeenCalledTimes(1)
```
maxammann: testing for calls of NativeFetcherModule.fetchAsync is actually testing implementation details.

I'd prefer to mock the result of `NativeFetcherModule.fetchAsync` and then asserting that `FetcherModule.fetchAsync` returns the correct result.

src/modules/fetcher/__mocks__/NativeFetcherModule.js
```diff
@@ -1,18 +1,25 @@
...
-  fetchAsync: (targetFilePaths: { [url: string]: string }) => Promise<FetchResultType>
+function mockFetchAsync (targetFilePaths: TargetFilePathsType): Promise<FetchResultType> {
+  const fetchResult: FetchResultType = {}
+  for (const path in targetFilePaths) {
+    fetchResult[path] = {
+      lastUpdate: moment.tz('20110204', 'UTC').toISOString(),
+      url: targetFilePaths[path],
+      errorMessage: null
+    }
+  }
```
Schedulaar: You can replace lines 8-15 using `mapValues` of lodash with sth like
```js
const fetchResult = mapValues(targetFilePaths, url => ({
      lastUpdate: moment.tz('20110204', 'UTC').toISOString(),
      url,
      errorMessage: null
}))
```

src/modules/fetcher/__mocks__/NativeFetcherModule.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+
+import type { FetchResultType, TargetFilePathsType } from '../FetcherModule'
+import { NativeEventEmitter } from 'react-native'
+import moment from 'moment-timezone'
+import { mapValues } from 'lodash/object'
+
+function mockFetchAsync (targetFilePaths: TargetFilePathsType): Promise<FetchResultType> {
+  const fetchResult = mapValues(targetFilePaths, url => ({
+    lastUpdate: moment.tz('20110204', 'UTC').toISOString(),
```
maxammann: You can simplty use `2004-06-14T23:34:30Z` here instead of using moment-timezone

src/modules/fetcher/__tests__/FetcherModule.spec.js
```diff
@@ -0,0 +1,48 @@
...
+import NativeFetcherModule from '../NativeFetcherModule'
+import type { TargetFilePathsType } from '../FetcherModule'
+
+jest.mock('../NativeFetcherModule')
+
+beforeEach(() => {
+  jest.clearAllMocks()
+})
+
+describe('FetcherModule', () => {
```
maxammann: A test of the progress callback is missing here.
To make sure the coverage is oke you can use the Coverage functionality in IntelliJ. You can find it next to the Run and Debug button. ("Run with coverage")

maxammann: You can create a jest.fn() mock and assert that it has been called for example a hundred times or that the final call had the return value of 1.

src/modules/fetcher/__tests__/FetcherModule.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  })
+  it('should return the fetch result data if fetchAsync call is valid', async () => {
+    const fetcherModule = new FetcherModule()
+    const fetchResult = await fetcherModule.fetchAsync(testTargetFilePaths, progress => console.log(progress))
+    expect(fetchResult).toMatchSnapshot()
+  })
+  it('should add eventType to listeners', async () => {
+    const fetcherModule = new FetcherModule()
+    await fetcherModule.fetchAsync(testTargetFilePaths, progress => console.log(progress))
+    expect(NativeFetcherModule.addListener).toHaveBeenCalledWith('progress')
```
maxammann: This asserts on implementation details. See the comment above.

klinzo: Not really testable, we agreed on just removing the test.

maxammann:
@Simple-codinger this is rdy for merge :)

PR #138, opened 2019-09-25
NATIVE-344: Fix language not available

klinzo:


src/modules/i18n/components/__tests__/I18nProvider.spec.js
```diff
@@ -53,15 +53,15 @@ describe('I18nProvider', () => {
...
     const appSettings = new AppSettings()
-    appSettings.setContentLanguage('de')
+    await appSettings.setContentLanguage('de')
     const mockSetContentLanguage = jest.fn()
 
     render(<I18nProvider setContentLanguage={mockSetContentLanguage} />)
 
     await waitForExpect(async () => {
-      expect(mockSetContentLanguage).not.toHaveBeenCalled()
+      expect(mockSetContentLanguage).toHaveBeenCalledWith('de')
```
Schedulaar: :+1: good catch.

src/routes/dashboard/containers/DashboardContainer.js
```diff
@@ -99,6 +99,10 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
     }
   }
 
+  if (route.status === 'loading') {
+    return { status: 'loading' }
+  }
```
Schedulaar: This seems like a hacky workaround to me. The route should only the status `loading` while there are network requests pending.

I came across this problem as well when working on the tests for the Dashboard and CategoriesContainer, and I fixed the bug there by adding a new status to the route.
I will propose a new PR that solves this (separately from the test-task)

PR #139, opened 2019-09-29
NATIVE-344: Fix language not available

Schedulaar:
This PR contains a fix for showing the languageNotAvailable message.
It's implemented using a new status for the categoriesRoute (this was necessary to have access on the availableLanguages even if the actual route is not able to load).

Other than that, I cleaned up the test for morphContentLanguage (replaced snapshots with cleaner inline-checks) and added peeking for events (since that was such a small issue and highly related to the files I touched anyways).
Furthermore I had to adjust the cityContentReducer such that it would return new state objects instead of mutated old state.


src/modules/endpoint/sagas/watchFetchCategory.js
```diff
@@ -30,32 +31,45 @@ function * fetchCategory (dataContainer: DataContainer, action: FetchCategoryAct
...
-      // Only get languages if we've loaded them, otherwise we're peeking
-      const cityLanguages = loadCriterion.shouldLoadLanguages() ? yield call(dataContainer.getLanguages, city) : []
-
       const push: PushCategoryActionType = {
         type: `PUSH_CATEGORY`,
         params: { categoriesMap, resourceCache, path, cityLanguages, depth, key, city, language }
       }
 
       yield put(push)
+    } else if (path === `/${city}/${language}`) {
```
maxammann: Why are you checking here for the root category? Shouldn't the PushCategoryLanguagesActionType also be put for other categories?

src/modules/endpoint/reducers/pushEvent.js
```diff
@@ -13,9 +13,7 @@ const pushEvent = (state: CityContentStateType, action: PushEventActionType): Ci
 
   const getEventRoute = (): EventRouteStateType => {
     if (!path) {
-      const allAvailableLanguages = new Map(cityLanguages.map(
-        language => [language.code, `/${city}/${language.code}/events`]
-      ))
+      const allAvailableLanguages = new Map(cityLanguages.map(lng => [lng.code, null]))
```
maxammann: I think the null is a bit odd here. This means that the list of events should be displayed right?

src/modules/app/StoreActionType.js
```diff
@@ -52,6 +53,15 @@ export type PushCategoryActionType = {|
     path: string, depth: number, key: string
   |}
 |}
+export type PushCategoryLanguagesActionType = {|
+  type: 'PUSH_CATEGORY_LANGUAGES', params: {|
```
maxammann: I think this name is a bit unclear as it also sets the status 'languagesNotAvailable'.

Maybe a name like 'CATEGORY_LANGAUGE_NOT_AVAILABLE'. Or something more generic 'SET_CATEGORY_STATUS'.

src/modules/endpoint/sagas/watchFetchEvent.js
```diff
@@ -1,42 +1,66 @@
...
-      // Only proceed if the content is ready to be pushed to the state. If not then the UI automatically displays an
-      // appropriate error
-      const [events, resourceCache, cityLanguages] = yield all([
+    const peeking = yield select(state => isPeekingRoute(state, { routeCity: city }))
+
+    const loadCriterion = new ContentLoadCriterion(criterion, peeking)
+    const languageValid = yield call(loadCityContent, dataContainer, city, language, loadCriterion)
+
+    // Only get languages if we've loaded them, otherwise we're peeking
+    const cityLanguages = loadCriterion.shouldLoadLanguages() ? yield call(dataContainer.getLanguages, city) : []
```
maxammann: This means we can close this issue, right? https://issues.integreat-app.de/browse/NATIVE-335

maxammann:
Sorry again didn't see the WIP 🙈

src/modules/app/StoreActionType.js
Schedulaar: Yeah, I'm also not quite happy with my naming here... 
But `CATEGORY_LANGUAGE_NOT_AVAILABLE` is not an action and `SET_CATEGORY_STATUS` is too generic. I might go with the first one though...

klinzo: Would it be possible to just use the `FetchCategoryFailedAction` with an additional param `languageNotAvailable` and set the route status accordingly? Language not  available leads to a failed fetch :man_shrugging:

src/modules/endpoint/reducers/cityContentReducer.js
```diff
@@ -27,61 +27,84 @@ export default (
...
+      }
     case 'MORPH_CONTENT_LANGUAGE':
       if (state === null) {
         throw Error('Cannot morph content language on not initialized cityContent')
       }
       return morphContentLanguage(state, action)
     case 'FETCH_EVENT': {
       const { language, path, key, city } = action.params
-      const newState = state === null ? createCityContent(city) : state
+      const newState = state === null ? createCityContent(city) : { ...state }
```
klinzo:
```suggestion
      const newState = state || createCityContent(city)
```

src/modules/endpoint/reducers/cityContentReducer.js
```diff
@@ -27,61 +27,84 @@ export default (
...
     }
     case 'CLEAR_EVENT': {
       const { key } = action.params
       if (state === null) {
-        return state
+        return null
       }
       delete state.eventsRouteMapping[key]
-      return state
+      return { ...state }
```
klinzo: Why did you change this?

src/modules/endpoint/reducers/pushEvent.js
klinzo: Not really intuitive imo, but should be fine. For me this looks more like the language is not available, even though it is called `allAvailableLanguages` so it should be fine.

src/modules/endpoint/sagas/__tests__/watchFetchCategory.spec.js
```diff
@@ -118,16 +108,16 @@ describe('watchFetchCategories', () => {
         .run()
     })
 
-    it('should do nothing if content can not be retrieved (e.g. language invalid) and not peeking', async () => {
-      const { dataContainer, initialPath } = await createDataContainer(city, language)
+    it('should put pushCategoryLanguages action if content can not be retrieved for root model', async () => {
```
klinzo: Rather `if root model is not available in the language`

src/modules/endpoint/sagas/watchFetchCategory.js
klinzo: Nope. If the fetch failed for some other reason, we throw an error right? `languageValid` is only `false` if the language not exists (for the root category). But if the language does not exist for the root category, it surely does not exist for any other category. Or am I missing something here?

klinzo: Nope, I don't think so. `loadCityContent` only returns false if the language is not available for the root category, otherwise it should throw. But if the language  is not available for the root category, it is surely not available in any other category. Or am I missing something here?

klinzo: Same should apply to events. If the language is not available for the events list, it is also not available for a single event

src/modules/error/hocs/withPayloadProvider.js
```diff
@@ -51,7 +50,7 @@ const withPayloadProvider = <S: { dispatch: Dispatch<StoreActionType> }, R> (
 
       refresh = () => {
         const props = this.props
-        if (props.status === 'routeNotInitialized' || props.status === 'loading') {
+        if (props.status === 'routeNotInitialized' || props.status === 'loading' || props.status === 'languageNotAvailable') {
```
klinzo: You could do `['routeNotInitialized', ..].includes(props.status)`. Not sure if it is better though.

src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -40,66 +40,57 @@ type StatePropsType = StatusPropsType<ContainerPropsType, RefreshPropsType>
...
-  const stateView = new CategoriesRouteStateView(route.path, route.models, route.children)
-  if (!languages.includes(route.language)) {
+  if (route.status === 'languageNotAvailable') {
     return {
       status: 'languageNotAvailable',
-      availableLanguages: languages,
-      cityCode: city,
-      refreshProps,
-      changeUnavailableLanguage: createChangeUnavailableLanguage(route.path, ownProps.navigation, city)
+      availableLanguages: languages.filter(lng => route.allAvailableLanguages.has(lng.code)),
```
klinzo: Why is this called `allAvailableLanguages` btw? Is there a reason for it? One could assume `availableLanguages` are already all available languages.

src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -40,66 +40,57 @@ type StatePropsType = StatusPropsType<ContainerPropsType, RefreshPropsType>
...
-const createChangeUnavailableLanguage = (path: string, navigation: NavigationScreenProp<*>, city: string) => (
+const createChangeUnavailableLanguage = (city: string) => (
   dispatch: Dispatch<StoreActionType>, newLanguage: string
 ) => {
   const switchContentLanguage: SwitchContentLanguageActionType = {
     type: 'SWITCH_CONTENT_LANGUAGE',
     params: { newLanguage, city }
   }
   dispatch(switchContentLanguage)
-  const navigateToCategory = createNavigateToCategory('Categories', dispatch, navigation)
```
klinzo: Nice :+1:

src/modules/endpoint/sagas/watchFetchCategory.js
Schedulaar: True. We definitely know, that the language is not available for a specific page, if `loadCityContent` reutrns false. But in that case we are only given the information `load /augsburg/invalid-language/some-page` where we cannot determine any available language. So the only sensible thing we can do, is to show an error, that the requested page is not available at all.

src/modules/endpoint/reducers/pushEvent.js
Schedulaar: Yeah, but it's consistent with the `eventRouteConfig`. We can consider changing it there as well, though, but I'd extract that to another task.

src/modules/endpoint/reducers/cityContentReducer.js
Schedulaar: Reducers mustn't mutate the parameters (see https://redux.js.org/basics/reducers#handling-actions).
I modified the types now, such that flow prevents stuff like that now.

src/modules/error/hocs/withPayloadProvider.js
Schedulaar: I guess, flow doesn't understand that...

src/routes/categories/containers/CategoriesContainer.js
Schedulaar: It's an implicit indicator that it also contains the current language (in contrast to CategoryModel.availableLanguages)

src/modules/endpoint/sagas/watchFetchCategory.js
klinzo: Makes sense, yes :)

src/modules/error/hocs/withPayloadProvider.js
klinzo: Ahh yeah okay...

src/modules/app/StoreActionType.js
Schedulaar: I used the FetchCategoryFailedAction (and corresponding for events) now.

src/modules/endpoint/reducers/__tests__/cityContentReducer.spec.js
```diff
@@ -233,6 +294,44 @@ describe('cityContentReducer', () => {
     })?.categoriesRouteMapping).toEqual({})
   })
 
+  it('should set route status to languageNotAvailable on FETCH_CATEGORY_FAILED if allAvailableLanguages is not null', () => {
```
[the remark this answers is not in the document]
Schedulaar: The difference is, that the one is for events and the other is for categories.

src/modules/endpoint/sagas/__mocks__/loadCityContent.js
```diff
@@ -10,10 +10,8 @@ const loadCityContent = function * loadCityContent (
   criterion: ContentLoadCriterion
 ): Saga<boolean> {
   const languagesAvailable = yield call(() => dataContainer.languagesAvailable(newCity))
-  const eventsAvailable = yield call(() => dataContainer.eventsAvailable(newCity, newLanguage))
-  const categoriesAvailable = yield call(() => dataContainer.categoriesAvailable(newCity, newLanguage))
 
-  if (!languagesAvailable || (!eventsAvailable && !categoriesAvailable)) {
+  if (!languagesAvailable) {
```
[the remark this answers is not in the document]
Schedulaar: Well, I want to mock `loadCityContent` in the case, when `languages` are available, but not the requested language (and therefore also no events nor categories).

PR #140, opened 2019-10-08
NATIVE-15: Preselect the nearest city on the landing route

SimonVortkamp:
Design is not final yet, maybe the button will be removed, the handling if no city is close might change and the headline might change to highlight close cities more.


klinzo:
Will you address the points mentioned in the description (e.g. design of the button and handling if no city is close) in a separate issue/PR or will you do this in this PR?

src/routes/landing/components/CitySelector.js
```diff
@@ -25,11 +26,18 @@ type PropsType = {|
...
+  theme: ThemeType,
+  currentLongitude: number | null,
+  currentLatitude: number | null,
+  renderLocationList: boolean
 |}
 
-const byName = (name: string) => {
-  return (city: CityModel) => city.name.toLowerCase().includes(name)
+const checkAliases = (a: CityModel, filterText: string): boolean => {
+  return Object.keys(a.aliases || {}).map(key => key.toLowerCase().includes(filterText.toLowerCase())).includes(true)
```
Schedulaar: You can use `some` instead of `map(...).includes(true)`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

src/routes/landing/components/CitySelector.js
```diff
@@ -45,19 +53,67 @@ const developmentCompare = (a: CityModel, b: CityModel) => {
...
+type StateType = {
+  currentLongitude: null | number,
+  currentLatitude: null | number
+}
+
+class CitySelector extends React.PureComponent<PropsType, StateType> {
+  state = {
+    currentLongitude: null,
+    currentLatitude: null
+  }
```
Schedulaar: Since you get `currentLongitude` and `currentLatitude` via props now, you can remove them from the state and also remove the componentDidMount function.

SimonVortkamp:
I will adress these issues in a different PR since I don't know when i will get the final design decisions yet

src/routes/landing/components/FilterableCitySelector.js
```diff
@@ -17,34 +19,70 @@ type PropsType = {
...
+        const currentLatitude = position.coords.latitude
+        this.setState({ currentLongitude: currentLongitude })
+        this.setState({ currentLatitude: currentLatitude })
+      },
+      error => {
+        alert(error.message)
+      },
+      { enableHighAccuracy: false, timeout: 20000, maximumAge: 1000 }
+    )
+    this.setState({ renderLocationList: true })
```
klinzo: I guess we only want to render it if the locating was successful, so you should probably move this into the upper arrow function to the other `setStates`

src/routes/landing/components/FilterableCitySelector.js
```diff
@@ -17,34 +19,70 @@ type PropsType = {
...
+        <View style={{ flexDirection: 'row', justifyContent: 'space-between' }} >
+          <SearchInput style={{ flexGrow: 1 }}
           filterText={filterText}
           onFilterTextChange={this.onFilterTextChange}
           placeholderText={t('searchCity')}
           spaceSearch={false}
           theme={theme} />
+          <TouchableOpacity onPress={this._onPressLocationButton} style={{ flexGrow: 0 }}>
+            <Image source={
+              require('../../../../assets/app-locationButton.png')} />
```
klinzo: Just import this normally at the top of the file with an import.

src/routes/landing/components/FilterableCitySelector.js
```diff
@@ -17,34 +19,70 @@ type PropsType = {
 }
 
 type StateType = {
-  filterText: string
+  filterText: string,
+  currentLongitude: number | null,
+  currentLatitude: number | null,
+  renderLocationList: boolean
```
Schedulaar: This bool is redundant: We should always render the location list, if `currentLongitude` and `currentLatitude` are available (aka non null).
You can remove it.

src/routes/landing/components/CitySelector.js
```diff
@@ -45,19 +52,58 @@ const developmentCompare = (a: CityModel, b: CityModel) => {
   return a.sortingName.localeCompare(b.sortingName, 'en-US')
 }
 
+const degreesToRadians = (deg: number): number => {
+  const degreesSemicircle = 180
+  return Math.PI * deg / degreesSemicircle
+}
+
+const degreeDifference = (longitude0: number, latitude0: number, longitude1: number, latitude1: number): number => {
```
Schedulaar: I'd rather call this `calculateDistance`

src/routes/landing/components/CitySelector.js
```diff
@@ -76,10 +122,45 @@ class CitySelector extends React.PureComponent<PropsType> {
...
+      return null
+    }
+    const currentLatitude = this.props.currentLatitude
+    const currentLongitude = this.props.currentLongitude
+    let cities = this.props.cities.filter(_city => _city.live)
+      .sort((a: CityModel, b: CityModel) => compareDistance(a, b, currentLongitude, currentLatitude))
+    const numberOfClosestCities = 3
+    cities = cities.slice(0, numberOfClosestCities)
+    const maximalDistance = 90
+    cities = cities.filter(_city => currentDistance(_city, currentLongitude, currentLatitude) < maximalDistance)
```
Schedulaar: It would be nicer to define `cities` as a `const` variable right away by chaining the array operations like so: 
```js
    const numberOfClosestCities = 3
    const maximalDistance = 90
    const cities = this.props.cities
      .filter(_city => _city.live)
      .sort((a: CityModel, b: CityModel) => compareDistance(a, b, currentLongitude, currentLatitude))
      .slice(0, numberOfClosestCities)
      .filter(_city => calculateDistanceToCity(_city, currentLongitude, currentLatitude) < maximalDistance)
```

src/routes/landing/components/CitySelector.js
```diff
@@ -76,10 +122,45 @@ class CitySelector extends React.PureComponent<PropsType> {
...
+    if (cities && cities.length) {
```
Schedulaar: Just check cities.length here: `if (cities.length > 0)`

src/routes/landing/components/CitySelector.js
```diff
@@ -76,10 +122,45 @@ class CitySelector extends React.PureComponent<PropsType> {
...
+        <CityGroup theme={this.props.theme}> Close </CityGroup>
+        {cities.map(city => <CityEntry
+          key={city.code}
+          city={city}
+          filterText={this.props.filterText}
+          navigateToDashboard={this.props.navigateToDashboard}
+          theme={this.props.theme} />)}
+      </View>
+    } else {
+      alert('There are no cities offering integreat close to you.')
```
Schedulaar: We shouldn't use `alert` in the render method.
Instead return `<CityGroup theme={this.props.theme}>No Integreat places nearby</CityGroup>` or similar.

src/routes/landing/components/CitySelector.js
```diff
@@ -45,19 +52,58 @@ const developmentCompare = (a: CityModel, b: CityModel) => {
...
+  // $FlowFixMe https://github.com/facebook/flow/issues/2221
+  const coordinates: Array<CoordinatesType> = Object.values(a.aliases || {})
+  coordinates.push({ longitude: a.longitude, latitude: a.latitude })
+  const distances: Array<number> = coordinates.map((coords: CoordinatesType): {| value: number |} | null => {
+    if (coords.longitude === null || coords.latitude === null) {
+      return null
+    }
+    const longitude1 = degreesToRadians(coords.longitude)
+    const latitude1 = degreesToRadians(coords.latitude)
+    return { value: degreeDifference(longitude0, latitude0, longitude1, latitude1) * earthRadius }
```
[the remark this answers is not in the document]
Schedulaar: yeah, that was me :D Flow understands `.filter(Boolean)` and adjusts the type of the resulting array such that it cannot contain `null`s any more.
I didn't see, that `null` was not possible anyway here, so @SimonVortkamp could replace lines 76-84 with something like:
``` js
  const distances: Array<number> = coordinates.map(coords => {
    const longitude1 = degreesToRadians(coords.longitude)
    const latitude1 = degreesToRadians(coords.latitude)
    return calculateDistance(longitude0, latitude0, longitude1, latitude1) * earthRadius
  })
  return Math.min(...distances)
```

src/routes/landing/components/CitySelector.js
```diff
@@ -45,19 +52,58 @@ const developmentCompare = (a: CityModel, b: CityModel) => {
...
+  }).filter(Boolean).map(x => x.value)
+  return Math.min(...distances)
+}
+
+const compareDistance = (a: CityModel, b: CityModel, longitude: number, latitude: number) => {
```
NKlug: `cityModel`

src/routes/landing/components/CitySelector.js
```diff
@@ -76,10 +122,45 @@ class CitySelector extends React.PureComponent<PropsType> {
...
 
+  renderListByLocation (): React.Node {
+    if (this.props.currentLatitude === null || this.props.currentLongitude === null) {
+      return null
+    }
+    const currentLatitude = this.props.currentLatitude
+    const currentLongitude = this.props.currentLongitude
+    let cities = this.props.cities.filter(_city => _city.live)
+      .sort((a: CityModel, b: CityModel) => compareDistance(a, b, currentLongitude, currentLatitude))
+    const numberOfClosestCities = 3
```
NKlug: Maybe move constants to the start of the file (e.g. `NUMBER_OF_CLOSEST_CITIES = 3`)

src/routes/landing/components/FilterableCitySelector.js
```diff
@@ -17,34 +19,70 @@ type PropsType = {
...
+  }
+
   render () {
     const { cities, t, theme } = this.props
     const filterText = this.state.filterText
 
     return (
       <>
-        <SearchInput
+        <View style={{ flexDirection: 'row', justifyContent: 'space-between' }} >
```
NKlug: You can use a styled component for this:
something like:
```
const MyView = styled.View`
flexDirection: row; 
justifyContent: space-between`
```
(with a better name ofc). And then just use `<MyView></MyView>` instead.

src/routes/landing/components/FilterableCitySelector.js
NKlug: If I remember correctly we use (or used to use) `FastImage` instead of the default ReactNative `Image`.

ios/Integreat/Info.plist
```diff
@@ -63,5 +63,6 @@
 	</array>
 	<key>UIViewControllerBasedStatusBarAppearance</key>
 	<false/>
+	<key>NSLocationWhenInUseUsageDescription</key>
```
maxammann: The value is missing here: https://developer.apple.com/documentation/bundleresources/information_property_list/nslocationwheninuseusagedescription

I can do this on mac tough

src/modules/endpoint/DatabaseConnector.js
```diff
@@ -60,7 +60,10 @@ type ContentCityJsonType = {|
   prefix: string,
   extrasEnabled: boolean,
   eventsEnabled: boolean,
-  sortingName: string
+  sortingName: string,
+  longitude: number | null,
+  latitude: number | null,
```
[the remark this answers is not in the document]
maxammann: ?number includes undefined afaik. As this is the JSON type it should stay like it is.

src/routes/landing/components/CitySelector.js
```diff
@@ -76,10 +122,45 @@ class CitySelector extends React.PureComponent<PropsType> {
...
+    cities = cities.slice(0, numberOfClosestCities)
+    const maximalDistance = 90
+    cities = cities.filter(_city => currentDistance(_city, currentLongitude, currentLatitude) < maximalDistance)
+    if (cities && cities.length) {
+      return <View>
+        <CityGroup theme={this.props.theme}> Close </CityGroup>
```
maxammann: This needs to be added to the locales.json. So you have to the the translate() function and define a key in the locales.json

maxammann: Also I think Nearby sounds better. What do the others think? @markl @klinzo @NKlug @Simple-codinger

src/routes/landing/components/FilterableCitySelector.js
```diff
@@ -17,34 +19,70 @@ type PropsType = {
...
+      currentLatitude: null,
+      renderLocationList: false
+    }
   }
 
   onFilterTextChange = (filterText: string) => this.setState({ filterText })
 
+  _onPressLocationButton = () => {
+    if (Platform.OS === 'ios') {
+      Geolocation.requestAuthorization()
```
maxammann: Why is this only needed on ios?

src/routes/landing/components/FilterableCitySelector.js
maxammann: Correct, we have a custom Image component. See the Thumbnail component for the proper usage.

maxammann:
As soon as you are ready, ping me and I will test it on iOS

src/routes/landing/components/FilterableCitySelector.js
SimonVortkamp: On Android the request happens automatically, when you use Geolocation.getCurrentPosition()

maxammann:
Build currently fails because of iOS, going to take care of it on friday this week.

Schedulaar:
This is now tested on iOS.


We should add a proper error message though if the user denies the permission:

Simulator Screen Shot - iPhone XS - 2019-10-30 at 12 39 45


This can be outsourced to a separate task. I'll create one.

maxammann:
(Last comment was written by me)

maxammann:
@Schedulaar can you kick me out of your account?

maxammann:
I logged you out here in the Senkelfabrik :P

src/routes/landing/components/FilterableCitySelector.js
```diff
@@ -43,48 +49,41 @@ class FilterableCitySelector extends React.Component<PropsType, StateType> {
...
-        this.setState({ currentLatitude: currentLatitude })
-      },
-      error => {
-        alert(error.message)
-      },
+      position => this.setState({
+        currentLongitude: position.coords.longitude,
+        currentLatitude: position.coords.latitude
+      }),
+      alert(this.props.t('locationError')),
```
maxammann: I think the lambda `error =>` is missing here. Else this is executed always.

maxammann: Minor: I think it is better to use the Alert API of RN: https://facebook.github.io/react-native/docs/alert

maxammann:
Here is the diff for the locales: https://github.com/Integreat/integreat-locales/compare/a9a83229e399ec73077498480dcc690e0e77c366%5E...a9a83229e399ec73077498480dcc690e0e77c366

Schedulaar:
Sorry for closing the PR, I hit the wrong button :sweat_smile: 
@maxammann We should not use "Nearby cities" since not every integreat-instance is a city (but also a Kaff or a Landkreis). So I'd prefer "Nearby" here. Maybe we can use "Nearby places", but I'm not entirely happy with that either.

ios/Integreat/de.lproj/InfoPlist.strings
```diff
@@ -0,0 +1,10 @@
+/* 
+  InfoPlist.strings
+  Integreat
+
+  Created by Max Ammann on 15.07.19.
+  Copyright © 2019 Facebook. All rights reserved.
+*/
+
+NSPhotoLibraryAddUsageDescription = "Integreat möchte Bilder in der Gallerie speichern";
+NSLocationWhenInUseUsageDescription = "Integreat möchte nahegelegene Orte finden, welche lokale Informationen bereitstellen";
```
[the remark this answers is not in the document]
maxammann: There is a task for including them in the integreat-locales project. I think an automated approach is too complicated for now (Xcode would have to modify itself :man_shrugging: ).

maxammann:
Oke you are right, but for now that is fine imo as we also have the strings "Ort suchen". So yeah "Nearby places" seems fine. In that case also check whether the term "city/Stadt" is used on the Landing page.

klinzo:
I'd suggest to change the german translation of `noNearbyPlaces` to `Keine Orte in der Nähe gefunden`.  The german `locationError` is missing a `,` and yes, the english `loactionError` is probably wrong.

maxammann:
@SimonVortkamp should I take a look at the locales?

ios/Integreat/Info.plist
Simple-codinger: The strings are located in InfoPlist.strings, I don't think that it's necessary to put it in Info.plist as well.

PR #142, opened 2019-10-09
NATIVE-307: Use external locales

maxammann:


maxammann:
This is now open for review

PR #143, opened 2019-10-09
NATIVE-345: Fix missing thumbnails

klinzo:


src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -79,7 +79,7 @@ export default function * loadCityContent (
 
     const fetchMap = resourceURLFinder.buildFetchMap(
       categoriesMap.toArray().concat(events),
-      (url, path, urlHash) => buildResourceFilePath(url, newCity, urlHash)
+      (url, urlHash) => buildResourceFilePath(url, newCity, urlHash)
```
Schedulaar: Hmmm... `buildResourceFilePath` needs 4 arguments; `path` is one of them...

src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -1,31 +1,37 @@
...
 import type { DataContainer } from '../DataContainer'
+import type { FileCacheStateType } from '../../app/StateType'
 
-export type PathType = string
-export type UrlType = string
-export type FilePathType = string
-export type HashType = string
-export type FetchMapType = { [filePath: FilePathType]: [UrlType, PathType, HashType] }
+export type FetchMapTargetType = { url: string, filePath: string, urlHash: string }
+export type FetchMapEntryType = Array<FetchMapTargetType>
```
Schedulaar: You might remove this type, since it's probably less readable with it (too many types; my brain's buffer is limited :sweat_smile: )

src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -1,31 +1,37 @@
...
+export type FetchMapType = { [path: string]: FetchMapEntryType }
```
Schedulaar:
```suggestion
export type FetchMapType = { [categoryPath: string]: FetchMapEntryType }
```

src/modules/endpoint/sagas/fetchResourceCache.js
```diff
@@ -35,19 +41,16 @@ export default function * fetchResourceCache (
...
+    const resourceCache = mapValues(fetchMap, fetchMapEntry =>
+      reduce<FetchMapEntryType, FileCacheStateType>(fetchMapEntry, (acc, fetchMapTarget: FetchMapTargetType) => {
+        const filePath = fetchMapTarget.filePath
         const downloadResult = successResults[filePath]
-        const [, , urlHash] = fetchMap[filePath]
 
         if (downloadResult) {
           acc[downloadResult.url] = {
             filePath,
             lastUpdate: downloadResult.lastUpdate,
```
Schedulaar: This is always a string, but should be of type `Moment` :thinking:

src/modules/endpoint/sagas/fetchResourceCache.js
klinzo: Not only categories but also events here

klinzo: It's used for categories and events

src/modules/endpoint/sagas/loadCityContent.js
klinzo: Not really? `export default (urlString: string, city: string, urlHash: string) => {
  return `${getResourceCacheFilesDirPath(city)}/${urlHash}${getExtensionWithDot(urlString)}`
}`

Schedulaar: Ok, I didn't see, that you're branching off of NATIVE-328.

klinzo:
The problem was that if a resource was used more than once (of different categories), in the fetchMap the first occurrence was overwritten with the second one.

maxammann:
But shoudn't that be oke? I mean if multiple categories have the same resource then there is no need to fetch it multiple times.

The local file path should still point to the correct one, doesnt it?

klinzo:
There is no need to fetch it multiple times but the linking from the category using the resource to the downloaded resource was lost, because the resource was always downloaded on demand.

PR #145, opened 2019-10-23
NATIVE-282: Handle offline language switch

klinzo:
Since netinfos `isInternetReachable()` is only available from v4 and a migration to androidx would be necessary to use it, I'm just using a workaround here.


src/modules/endpoint/sagas/watchContentLanguageSwitch.js
```diff
@@ -47,6 +52,9 @@ export function * switchContentLanguage (dataContainer: DataContainer, action: S
     }
     yield put(insert)
   } catch (e) {
+    if (e.message === 'Network request failed') {
```
maxammann: We should add here a note into some issue to change this. I added it here: https://issues.integreat-app.de/browse/NATIVE-354

src/routes/dashboard/containers/DashboardContainer.js
```diff
@@ -133,11 +134,16 @@ const DashboardContainer = (props: ContainerPropsType) => {
...
 
-export default withRouteCleaner<PropsType>(
-  connect<PropsType, OwnPropsType, _, _, _, _>(mapStateToProps, mapDispatchToProps)(
-    omitNavigation<PropsType>(
-      withPayloadProvider<ContainerPropsType, RefreshPropsType>(refresh)(
-        DashboardContainer
-      )
-    )
-  ))
+type RestType = $Diff<PropsType, OwnPropsType>
```
maxammann: Did this fix some flow error?

src/modules/endpoint/sagas/watchContentLanguageSwitch.js
```diff
@@ -47,6 +52,9 @@ export function * switchContentLanguage (dataContainer: DataContainer, action: S
...
+      Alert.alert(t('languageSwitchFailedTitle'), t('languageSwitchFailedMessage'))
```
maxammann: I think alert is not really a good way to indicate a failure.

Best would be if the components react to this.

klinzo: In this special case I think it's fine, because we don't really have a component which could react to the failure. The ChangeLanguageModel just dispatches the action and gets closed immediately after, and the page displayed before is also not responsible imo.

src/routes/dashboard/containers/DashboardContainer.js
klinzo: This is used to remove `t` and `navigation` from props (passed through `ownProps`). Before just `navigation` was removed via `omitNavigation`

src/modules/endpoint/sagas/watchContentLanguageSwitch.js
maxammann: This is basically like a `languageNotAvailable` error. So this should somehow modify the state and replace the content with an error.

This can be done later though.

klinzo: No, it's not. LanguageNotAvailable means the language is not used in the city, e.g. lk augsburg has just `de` and `en`, so visiting it with `fr` produces a `languageNotAvailable` error

klinzo: If this should be done otherwise, it is should be done post mortem.

klinzo:
@Schedulaar ping

src/modules/app/StoreActionType.js
```diff
@@ -109,7 +110,7 @@ export type EventsActionType =
 
 export type SwitchContentLanguageActionType = {|
   type: 'SWITCH_CONTENT_LANGUAGE', +params: {|
-    +newLanguage: string, +city: string
+    +newLanguage: string, +city: string, +t: TFunction
```
Schedulaar: I'd like to avoid having functions in actions. Is there another way to give the sagas the translate function? For example by setting up the i18n before the redux store; That way we could also initialize the content language in the store directly with the information provided by i18n. Also you would have to touch a lot less files

src/modules/endpoint/DataContainer.js
```diff
@@ -87,6 +87,11 @@ export interface DataContainer {
    */
   eventsAvailable (city: string, language: string): Promise<boolean>,
 
+  /**
+   * Returns whether the city content (e.g. events, categories, ...) have been loaded or not.
+   */
+  cityContentAvailable (city: string, language: string): Promise<boolean>,
```
Schedulaar: What happens for city/language pairs that have only been peeked?
Maybe add this to the comment above

src/modules/endpoint/DefaultDataContainer.js
```diff
@@ -169,25 +169,29 @@ class DefaultDataContainer implements DataContainer {
...
   }
 
-  async eventsAvailable (city: string, language: string): Promise<boolean> {
+  eventsAvailable = async (city: string, language: string): Promise<boolean> => {
     const context = new DatabaseContext(city, language)
     return this.isCached('events', context) || this._databaseConnector.isEventsPersisted(context)
   }
+
+  cityContentAvailable = async (city: string, language: string): Promise<boolean> => {
+    return this.categoriesAvailable(city, language) && this.eventsAvailable(city, language)
```
Schedulaar: what about `languages`?

src/modules/endpoint/sagas/watchContentLanguageSwitch.js
klinzo: Task created

src/modules/app/StoreActionType.js
klinzo: Since the alert will be replaced by the error component in https://issues.integreat-app.de/browse/NATIVE-359  ,hence the TFunction will not be required anymore anyway. Therefore we agreed on keeping this at it is.

PR #147, opened 2019-10-24
NATIVE-300: Add timeout before showing loading spinner

Schedulaar:


Schedulaar:
It's just a workaround for a bug that seems to occur for iOS quite unexpectedly when you initialize the view with an active RefreshControl. Since we wanted to show it only once loading is slow (> 800ms), we tried using a delay before displaying the loading spinner which solved the issue :D

PR #150, opened 2019-10-30
NATIVE-323: Test Dashboard and Categories containers

Schedulaar:


src/routes/categories/containers/__tests__/CategoriesContainer.spec.js
```diff
@@ -0,0 +1,251 @@
...
+      navigation,
+      resourceCache,
+      stateView: expect.any(CategoriesRouteStateView),
+      t: expect.any(Function),
+      theme: brightTheme
+    })
+    const stateView = categoriesInstance.props.stateView
+    expect([stateView.root(), stateView.children()]).toEqual([rootCategory, categoriesMap.getChildren(rootCategory)])
+  })
+})
```
klinzo: Nicely done :+1:

src/routes/dashboard/containers/__tests__/DashboardContainer.spec.js
```diff
@@ -0,0 +1,254 @@
...
+      navigation,
+      resourceCache,
+      stateView: expect.any(CategoriesRouteStateView),
+      t: expect.any(Function),
+      theme: brightTheme
+    })
+    const stateView = dashboardInstance.props.stateView
+    expect([stateView.root(), stateView.children()]).toEqual([rootCategory, categoriesMap.getChildren(rootCategory)])
+  })
+})
```
klinzo: Much of duplicated code between `CategoriesContainer` and `DashboardContainer` as well as between `CategoriesContainer.spec` and `DashboardContainer.spec`, maybe come code parts could be  reused. Probably  not worth the effort though.

src/modules/app/StateType.js
```diff
@@ -75,14 +75,18 @@ export type FileCacheStateType = $ReadOnly<{
...
 
-export type ErrorStateType = {|
-  +errorMessage: string
-|}
-
 export type LanguageResourceCacheStateType = $ReadOnly<{
   [path: string]: FileCacheStateType
 }>
 
+export type CityContentResourceCacheStateType = {|
```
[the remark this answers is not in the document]
maxammann: We could just name it `ResourceCacheStateType` There is no need to put the CityContent there afaik see

src/routes/dashboard/containers/__tests__/DashboardContainer.spec.js
```diff
@@ -0,0 +1,254 @@
...
+        eventsRouteMapping: {},
+        resourceCache: resourceCacheState || { status: 'ready', value: resourceCache },
+        searchRoute: null
+      },
+      contentLanguage: 'de',
+      cities: cities || { status: 'ready', models: [city] }
+    }
+  }
+
+  const rootCategory = categoriesMap.findCategoryByPath(`/${city.code}/${language.code}`)
```
maxammann: You can also find the root category by finding the category where `category.root == true`

src/routes/dashboard/containers/__tests__/DashboardContainer.spec.js
```diff
@@ -0,0 +1,254 @@
...
+  const models = reduce(
+    flatMap(
```
maxammann: There is also a flatMap from the std lib. Not sure whether it works yet though: https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

src/routes/dashboard/containers/__tests__/DashboardContainer.spec.js
```diff
@@ -0,0 +1,254 @@
...
+        resourceCache: resourceCacheState || { status: 'ready', value: resourceCache },
+        searchRoute: null
+      },
+      contentLanguage: 'de',
+      cities: cities || { status: 'ready', models: [city] }
+    }
+  }
+
+  const rootCategory = categoriesMap.findCategoryByPath(`/${city.code}/${language.code}`)
+  const models = reduce(
```
maxammann: If I get this correctly then models contains the models 2 levels deep. Isn't it just possible to use categoriesMap.toArray()?

If not then this should be extracted into a separate function. This code is quite complicated.

src/routes/categories/containers/__tests__/CategoriesContainer.spec.js
```diff
@@ -0,0 +1,251 @@
...
+  const expectLoadingIndicator = (state: StateType) => {
+    const store = mockStore(state)
+    const navigation = createNavigationScreenPropMock()
+    navigation.state.key = 'route-id-0'
+    jest.doMock('../../../../modules/categories/components/Categories', () => MockCategories)
+    const CategoriesContainer = require('../CategoriesContainer').default
+    const result = TestRenderer.create(
+      <Provider store={store}><CategoriesContainer navigation={navigation} /></Provider>
+    )
+    jest.advanceTimersByTime(1000)
```
maxammann: It would be good to have this constant depend on the 800ms delay. At least this should be extracted to a constant.

src/routes/categories/containers/__tests__/CategoriesContainer.spec.js
```diff
@@ -0,0 +1,251 @@
...
+        resourceCache: resourceCacheState || { status: 'ready', value: resourceCache },
+        searchRoute: null
+      },
+      contentLanguage: 'de',
+      cities: cities || { status: 'ready', models: [city] }
+    }
+  }
+
+  const rootCategory = categoriesMap.findCategoryByPath(`/${city.code}/${language.code}`)
+  const models = reduce(
```
maxammann: Same as in DashboardContainer test

src/routes/categories/containers/__tests__/CategoriesContainer.spec.js
```diff
@@ -0,0 +1,251 @@
...
+        eventsRouteMapping: {},
+        resourceCache: resourceCacheState || { status: 'ready', value: resourceCache },
+        searchRoute: null
+      },
+      contentLanguage: 'de',
+      cities: cities || { status: 'ready', models: [city] }
+    }
+  }
+
+  const rootCategory = categoriesMap.findCategoryByPath(`/${city.code}/${language.code}`)
```
maxammann: Same as in DashboardContainer test

src/modules/endpoint/reducers/pushCategory.js
```diff
@@ -32,8 +32,9 @@ const pushCategory = (state: CityContentStateType, action: PushCategoryActionTyp
   })
 
   // If there is an error in the old resourceCache, we want to override it
```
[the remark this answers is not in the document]
Schedulaar: No, it's still valid. If there's an error, then state.resourceCache.status will be 'error' and hence, we will override it with the new resourceCache without adding anything (because there's nothing to add sth to)

src/routes/dashboard/containers/__tests__/DashboardContainer.spec.js
Schedulaar: Yeah, but that's because right now, the Dashboard's pretty much the same as the Categories route. Since this will change in the future, it makes no sense to reduce the duplicity here imo...

src/routes/dashboard/containers/__tests__/DashboardContainer.spec.js
Schedulaar: It doesn't work unfortunately. It's too experimental :D

src/routes/dashboard/containers/__tests__/DashboardContainer.spec.js
Schedulaar: Seems more complicated to me.

src/routes/dashboard/containers/__tests__/DashboardContainer.spec.js
Schedulaar: I've simplified it.

PR #153, opened 2019-11-03
NATIVE-267: Cleanup settings screen

Simple-codinger:


src/modules/platform/constants/NativeConstants.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+
+import { NativeModules } from 'react-native'
+
+type NativeConstantsType = {
```
maxammann: I think it is good to export this type

src/modules/platform/constants/NativeConstants.js
```diff
@@ -0,0 +1,15 @@
+// @flow
```
maxammann: If you want you can create a "native-constants" module in the `src/modules` directory as it is basically a standalone module.

src/modules/layout/components/EmptyHeader.js
```diff
@@ -0,0 +1,66 @@
...
+        <HorizontalLeft>
+          <HeaderBackButton onPress={this.goBack} />
+          <HeaderText theme={this.props.theme}>{this.props.t('settings')}</HeaderText>
+        </HorizontalLeft>
+      </Horizontal>
+    </BoxShadow>
+  }
+}
+
+export default EmptyHeader
```
Schedulaar: Rename to `SettingsHeader`

PR #154, opened 2019-11-04
NATIVE-321: Fix displaying of events

NKlug:
There were two problems:
1. If `renderHtml` received an empty `files` object, it crashed. Now this is prevented.
2. The reason `renderHtml` received an empty `files` object was that when scanning the remoteContent for URLs to replace, old appearances of the same URLs were overwritten (this happens e.g. when two pages use the same image/icon). ~I also fixed this.~ Someone else (maybe without knowing) fixed this already.


NKlug:
This means that right now, when multiple pages in the same city include the same resource, we only download and store it once. This should not be a problem as long as we don't start deleting single files for a city. (Right know we only delete all files for the city, at least that's what I take out of NATIVE-266 / #151 )

klinzo:
Was fixed in https://issues.integreat-app.de/browse/NATIVE-345.

src/modules/common/renderHtml.js
```diff
@@ -10,6 +10,11 @@ const renderJS = (files: PageResourceCacheStateType) => `
   var hrefs = document.querySelectorAll('[href]')
   var srcs = document.querySelectorAll('[src]')
   var files = ${JSON.stringify(files)}
+    
+    if (!files) {
+      console.warn('Files for WebView undefined! Not replacing urls...' )
```
klinzo: Is it just a problem if undefined? Or also if files are empty? If the latter, maybe mention it.

NKlug: It's not a problem if the files are empty. But in this case the remaining part of the function has no effect, so we might just as well `return` in this case.
Do you want me to change the error message?

maxammann: What would happen here if you use the js api: alert(...)? :thinking: 

It would be cool if a react-native or some other alert would pop up if errors occur. Maybe also surround everyting with a try catch to propagate errors to the UI.

Right now this warning is basically useless as no one (not even all developers) can see this message.

maxammann: There is this stone-age issue: https://issues.integreat-app.de/browse/NATIVE-9

If the alert would give feedback to the user then we can close this I think :)

Schedulaar: If it's empty, it's not necessarily a faulty state, is it?

NKlug: It's not. But in this case the remaining lines of the function have 0 effect.

Schedulaar: Right, but no need to do anything special here

NKlug: I'll try. If it's not working, I can also just as well forward all errors (and therefore complete NATIVE-9). Shouldn't be that much of a hassle.

NKlug: So you suggest to check for `files === undefined` instead, right?

Schedulaar: No, I suggest not to change it :D `!files` is only true, if `files` is `undefined` or `null`. If `files` is an empty object, it evaluates to `true`.

NKlug: Ok nvm. I think we both mean the same thing then :D

klinzo: Yeah, me too :D

maxammann: If it does not work right away then skip this.

We'll prioritise NATIVE-9 if we think we need it.

NKlug: Ok. `alert()` works. Do we want to display an error message that is useful for the end user? Or just show one if we are in debug mode and target the developers?

maxammann:
How is the altert looking compared to native android/ios alerts?

NKlug:
To me it looks the same. I can show you tomorrow (Friday).

maxammann:
The `files` parameter can never be null/undefined. This is due to the following function: 
```
 getCategoryResourceCache (category: CategoryModel): PageResourceCacheStateType {
    return this.props.resourceCache[category.path] || {}
  }
```

The returned value is either the non-null value of an empty object.

PR #155, opened 2019-11-06
NATIVE-365: Make submodule url relative

maxammann:
In order to apply these changes you have to do: git submodule resync

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "locales"]
 	path = locales
-	url = https://github.com/Integreat/integreat-locales.git
+	url = ../integreat-locales.git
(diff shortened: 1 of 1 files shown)


Schedulaar:
Revert "Change head of locales to master". This is part of NATIVE-15. Also it's no use, since `git update` only checks out a specific commit, never a branch, so it doesn't resolve the issue with the detached HEAD (https://git-scm.com/book/de/v1/Git-Tools-Submodule#H%C3%A4ufige-Probleme-mit-Submodulen)

maxammann:
Reverted, I added update = merge tough. This should not detach the head for future updates.

maxammann:
@Schedulaar @NKlug @Simple-codinger ping

PR #156, opened 2019-11-06
NATIVE-138: Add adaptive icons

maxammann:


maxammann:
I used the no-round icon for generating the app icon. I included it because it is hard to find in the drive.

PR #158, opened 2019-11-11
NATIVE-355: Don't throw if city content is null

Simple-codinger:


src/modules/endpoint/reducers/cityContentReducer.js
```diff
@@ -30,12 +30,14 @@ export default (
       return { ...state, languages: action.params.languages }
     case 'PUSH_CATEGORY':
       if (state === null) {
-        throw Error('Cannot push category on not initialized cityContent')
+        // throw Error('Cannot push category on not initialized cityContent')
```
maxammann: This comment should be removed or replaced with a proper comment.

src/modules/endpoint/reducers/cityContentReducer.js
```diff
@@ -30,12 +30,14 @@ export default (
...
+        return null
```
maxammann: I'd prefere to avoid setting the `state` to `null` instead of expecting the state to be null.

Also you can remove the whole `if` here. If the state is already null then you do not need to do anything as the default case will return the `state` (which is null)

src/modules/endpoint/reducers/__tests__/cityContentReducer.spec.js
```diff
@@ -85,7 +85,11 @@ describe('cityContentReducer', () => {
   ]
   for (const action of unsupportedActionsOnUnitializedState) {
     it(`should throw on ${action.type} if state is unitialized`, () => {
-      expect(() => cityContentReducer(null, action)).toThrow()
+      if (action === pushCategoryAction || action === pushEventAction) {
+        expect(cityContentReducer(null, action)).toBeNull()
```
Schedulaar: Shouldn't the same apply for most actions that threw errors before? So for example also for morphContentLanguage, etc.? @maxammann

maxammann: Not sure what you're referencing here. Do you  mean the test should be similar for other reducers?

klinzo: I think he means that there are still some actions for which the cityContentReducer throws if the state is null. This should probably also be changed to just return null.

src/modules/endpoint/reducers/cityContentReducer.js
maxammann: Not sure about a better way here.

src/modules/endpoint/reducers/cityContentReducer.js
```diff
@@ -30,17 +30,18 @@ export default (
...
       if (state === null) {
-        throw Error('Cannot push event on not initialized cityContent')
+        return null
       }
       return pushEvent(state, action)
     case 'MORPH_CONTENT_LANGUAGE':
       if (state === null) {
-        throw Error('Cannot morph content language on not initialized cityContent')
+        return null
       }
```
Schedulaar: The same should apply to `FETCH_CATEGORY_FAILED`, `FETCH_RESOURCES_FAILED` and probably also to `SWITCH_CONTENT_LANGUAGE_FAILED`.

src/modules/endpoint/reducers/__tests__/cityContentReducer.spec.js
```diff
@@ -85,7 +85,12 @@ describe('cityContentReducer', () => {
   ]
   for (const action of unsupportedActionsOnUnitializedState) {
     it(`should throw on ${action.type} if state is unitialized`, () => {
```
Schedulaar: Probably the test should be split up into the ones that throw and the ones that return null.
Then you can name the tests according to what the actions do.

PR #161, opened 2019-11-13
NATIVE-336: Recover from malformatted JSON

Schedulaar:


src/modules/endpoint/sagas/__tests__/loadEvents.spec.js
```diff
@@ -66,4 +68,14 @@ describe('loadEvents', () => {
 
     expect(await dataContainer.getEvents(city, language)).toBe(otherEvents)
   })
+
+  it('should fetch events if the stored JSON is malformatted', async () => {
+    const context = new DatabaseContext('augsburg', 'de')
+    const path = new DatabaseConnector().getContentPath('events', context)
+    await RNFetchBlob.fs.writeFile(path, `{ "i": { "am": "malformatted" } }`, 'utf-8')
+    const dataContainer = new DefaultDataContainer()
+    const events = await runSaga({}, loadEvents, city, language, dataContainer, false).toPromise()
```
maxammann: Why does this not throw an exception? I can not see that there is a new try-catch added in the DatabaseConnector.js.
Only for lastUpdate.

Schedulaar: Please have a look at the changes in `loadEvents`.

src/modules/endpoint/DatabaseConnector.js
```diff
@@ -146,6 +156,9 @@ class DatabaseConnector {
     return `${CACHE_DIR_PATH}/cities.json`
   }
 
+  /**
+   * Prior to storing lastUpdate, there needs to be a lastUsage of the city in the context.
```
klinzo:
```suggestion
   * Prior to storing lastUpdate, the lastUsage of the city has to be set.
```
Context is misleading here imo.

src/modules/endpoint/DatabaseConnector.js
```diff
@@ -197,25 +198,49 @@ class DatabaseConnector {
...
+  }
 
-    const { last_update: lastUpdate } = metaData[cityCode]?.languages[languageCode] || {}
-    return lastUpdate ? moment(lastUpdate, moment.ISO_8601) : null
+  async _storeMetaCities (metaCities: MetaCitiesType) {
+    const path = this.getMetaCitiesPath()
+    const citiesMetaJson: MetaCitiesJsonType = mapValues(metaCities, cityMeta => ({
+      languages: mapValues(
+        cityMeta.languages,
+        ({ lastUpdate }): { last_update: string } => ({ last_update: lastUpdate.toISOString() })
```
klinzo: lastUpdate^4 :D

PR #163, opened 2019-11-13
NATIVE-329: Only cache files from CMS

Schedulaar:


src/modules/endpoint/constants.js
```diff
@@ -2,3 +2,4 @@
 
 export const baseUrl = 'https://cms-test.integreat-app.de'
 export const wohnenApiBaseUrl = 'https://api.wohnen.integreat-app.de/v0'
+export const allowedResourceHostNames = ['cms.integreat-app.de', 'cms-test.integreat-app.de']
```
maxammann: @svenseeberg are these the only hostnames which we use? Are there still links using web.integreat-app.de?

svenseeberg: Yes, for now these are the only domains.

PR #165, opened 2019-11-21
NATIVE-161: Rename extras to offers

SimonVortkamp:


klinzo:
Are there no other locales containing the term `extras`?

klinzo:
@maxammann are there other locales which should be changed? or is this good to go?

klinzo:
Alright, not quite sure what to do here: With this changes we have the keys `offers` and `extras` mixed. Since the locales have already been updated by @maxammann prior to this task with the correct terms `Offer`, I'd reject this and rename all the keys in a different task (or don't rename at all).

maxammann:
This is only a small part of the change. This is required right now such that the locales.json is compatible with the current source code. The other stuff will be done in: https://issues.integreat-app.de/browse/WEBAPP-497

PR #166, opened 2019-11-22
NATIVE-284: Fix wrong alignment of thumbnail in first level

Schedulaar:


Schedulaar:
The text is still centered.

PR #171, opened 2019-11-23
NATIVE-193: Add error codes

Schedulaar:
This introduces error codes and the ability to display these error codes when the occur in the PayloadProvider.

I had to introduce some changes to the integreat-api-client. See https://github.com/Integreat/integreat-api-client/pull/18 for the corresponding PR.


src/modules/error/ErrorCode.js
```diff
@@ -0,0 +1,32 @@
...
+
+type ErrorCodesType = {|
+  PageNotFound: 'pageNotFound',
+  NetworkConnectionFailed: 'networkConnectionFailed',
+  NetworkRequestFailed: 'networkRequestFailed', // server's http-status was not 200
+  ResponseMappingFailed: 'responseMappingFailed',
+  UnknownError: 'unknownError'
+|}
+
+export const ErrorCodes: ErrorCodesType = {
```
maxammann: I think this file should be called `ErrorCodes` and should export this as default.

PR #172, opened 2019-12-02
NATIVE-7: AsyncStorageVersion

Simple-codinger:


src/modules/settings/AppSettings.js
```diff
@@ -71,6 +80,12 @@ class AppSettings {
   clearSelectedCity = async () => {
     await this.asyncStorage.removeItem(SELECTED_CITY_KEY)
   }
+
+  clearAppSettings = async () => {
+    await this.asyncStorage.removeItem(CONTENT_LANGUAGE_KEY)
+    await this.asyncStorage.removeItem(SELECTED_CITY_KEY)
+    await this.asyncStorage.removeItem(ASYNC_STORAGE_VERSION_KEY)
```
Schedulaar: Should we use `this.asyncStorage.clear()` here?

Simple-codinger: Maybe in the future there will be some settings that we don't want to delete.

PR #174, opened 2019-12-18
NATIVE-398: Don't show null

klinzo:


src/routes/events/components/Events.js
```diff
@@ -109,8 +109,9 @@ class Events extends React.Component<PropsType> {
           <>
             <PageDetail identifier={t('date', { lng: language })} information={event.date.toFormattedString(language)}
                           theme={theme} language={language} />
-            <PageDetail identifier={t('location', { lng: language })} information={event.location.location}
-                          theme={theme} language={language} />
+            {event.location.location && <PageDetail identifier={t('location', { lng: language })}
```
klinzo: Why do we need the `{ lng: language }` config here? @maxammann @Schedulaar

maxammann: Can be removed :+1:

PR #181, opened 2020-01-15
NATIVE-408: Fix endless loading

klinzo:


src/modules/layout/components/Header.js
```diff
@@ -108,17 +109,13 @@ class Header extends React.PureComponent<PropsType> {
...
-  isPeeking (): boolean {
-    return this.props.peeking
-  }
-
   onShare = async () => {
     const { navigation, t } = this.props
     const sharePath: ?string = navigation.getParam('sharePath')
     if (!sharePath) {
-      return console.error('sharePath is undefined')
+      console.error('sharePath is undefined')
```
Schedulaar: I guess, we should only show the `Share` Menu-option, if there's a specific `sharePath` given in the navigation params. (a general integreat link doesn't make sense here imo).

src/modules/layout/components/Header.js
```diff
@@ -171,9 +168,10 @@ class Header extends React.PureComponent<PropsType> {
...
         </HorizontalLeft>
         <MaterialHeaderButtons>
-          {this.renderItem('Search', 'search', 'always', !this.isPeeking() ? this.goToSearch : undefined)}
-          {this.renderItem('Change Language', 'language', 'always',
-            !this.isPeeking() && goToLanguageChange ? goToLanguageChange : undefined)}
+          {!peeking && categoriesAvailable &&
+            this.renderItem('Search', 'search', 'always', this.goToSearch)}
+          {!peeking && goToLanguageChange &&
+          this.renderItem('Change Language', 'language', 'always', goToLanguageChange)}
           {this.renderItem(t('share'), undefined, 'never', sharePath ? this.onShare : undefined)}
```
Schedulaar: you might just show this only if sharePath is defined

src/modules/layout/components/Header.js
```diff
@@ -171,9 +168,10 @@ class Header extends React.PureComponent<PropsType> {
           {cityModel && <HeaderText theme={theme}>{this.cityDisplayName(cityModel)}</HeaderText>}
         </HorizontalLeft>
         <MaterialHeaderButtons>
-          {this.renderItem('Search', 'search', 'always', !this.isPeeking() ? this.goToSearch : undefined)}
-          {this.renderItem('Change Language', 'language', 'always',
-            !this.isPeeking() && goToLanguageChange ? goToLanguageChange : undefined)}
+          {!peeking && categoriesAvailable &&
+            this.renderItem('Search', 'search', 'always', this.goToSearch)}
```
Schedulaar: Why do we hide the search, when categories are not available? Does that make the header glitch? (we should avoid glitches)

src/routes/dashboard/containers/DashboardContainer.js
```diff
@@ -70,14 +70,18 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
 
-  if (state.cities.status === 'loading' || switchingLanguage || route.status === 'loading' || !languages) {
+  if (state.cities.status === 'loading' || switchingLanguage || route.status === 'loading' ||
+    languages.status === 'loading') {
     return { status: 'loading' }
   }
 
   if (route.status === 'languageNotAvailable') {
+    if (languages.status !== 'ready') {
+      throw Error('languageNotAvailable status impossible if languages not ready')
```
Schedulaar: We should never throw here! Rather return a general error or something different.

src/routes/events/containers/EventsContainer.js
```diff
@@ -61,14 +61,18 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
 
-  if (state.cities.status === 'loading' || switchingLanguage || route.status === 'loading' || !languages) {
+  if (state.cities.status === 'loading' || switchingLanguage || route.status === 'loading' ||
+    languages.status === 'loading') {
     return { status: 'loading' }
   }
 
   if (route.status === 'languageNotAvailable') {
+    if (languages.status !== 'ready') {
+      throw Error('languageNotAvailable status impossible if languages not ready')
```
Schedulaar: We should never throw here!

src/modules/layout/components/Header.js
klinzo: Why not? Just recommending integreat isn't that bad, is it?

src/modules/layout/components/Header.js
klinzo: Because search is a search for categories and it does not make sense to show it if categories are not available. If there are no categories, a loading spinner is displayed. This doesn't make sense if the fetching of categories failed (or they are still loading).

klinzo: Also, the header kind of glitches anyway, since only when the categories are pushed, the city name is displayed in the header. And I really think it looks good :) But test it yourself

src/modules/layout/components/Header.js
maxammann: https://issues.integreat-app.de/browse/NATIVE-417

src/modules/app/StateType.js
```diff
@@ -31,9 +31,7 @@ export type CategoryRouteStateType = {|
   +children: $ReadOnly<{ [path: PathType]: $ReadOnlyArray<PathType> }>
 |} | {|
   +status: 'languageNotAvailable',
-  +depth: number,
-  +city: string,
-  +language: string,
+  ...CategoryRouteConfigType,
```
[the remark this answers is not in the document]
klinzo: Yes, we need it for the `refreshProps` in the containers if the language is not available

src/modules/layout/components/Header.js
klinzo: Reverted

klinzo:
@Schedulaar

src/modules/layout/components/Header.js
maxammann: I just tested this and it looks pretty nice. The missing content in the header just plops into place. No glitching or something.

src/routes/categories/containers/CategoriesContainer.js
```diff
@@ -67,9 +68,17 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
 
+  const refreshProps = {
+    cityCode: route.city,
+    language: route.language,
+    path: route.path,
+    navigation: ownProps.navigation
+  }
+
   if (route.status === 'languageNotAvailable') {
     if (languages.status !== 'ready') {
```
Schedulaar: Can't you check for `languages.status === 'error'` and then use the `languages.message` as error message and code?

src/routes/dashboard/containers/DashboardContainer.js
```diff
@@ -75,9 +76,17 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
+    language: route.language,
+    path: route.path,
+    navigation: ownProps.navigation
+  }
+
   if (route.status === 'languageNotAvailable') {
     if (languages.status !== 'ready') {
-      throw Error('languageNotAvailable status impossible if languages not ready')
+      console.error('languageNotAvailable status impossible if languages not ready')
+      return { status: 'error', refreshProps, code: ErrorCodes.UnknownError, message: null }
```
Schedulaar: Same here

src/routes/events/containers/EventsContainer.js
```diff
@@ -66,9 +67,17 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
...
+    cityCode: route.city,
+    language: route.language,
+    navigation: ownProps.navigation
+  }
+
   if (route.status === 'languageNotAvailable') {
     if (languages.status !== 'ready') {
-      throw Error('languageNotAvailable status impossible if languages not ready')
+      console.error('languageNotAvailable status impossible if languages not ready')
+      return { status: 'error', refreshProps, code: ErrorCodes.UnknownError, message: null }
```
Schedulaar: Same here

src/modules/app/StateType.js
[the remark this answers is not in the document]
Schedulaar: It can happen, that we don't have a path! For example if you change the language of a subsubscreen and the screen doesn't have the targeted language available. Then there's no path that we can set here.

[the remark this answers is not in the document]
Schedulaar: So it was actually on purpose, that we cannot pull-to-refresh on a languageNotAvailable screen

src/modules/endpoint/reducers/morphContentLanguage.js
```diff
@@ -12,13 +12,14 @@ const categoryRouteTranslator = (newCategoriesMap: CategoriesMapModel, city: str
...
     }
-    const { depth, allAvailableLanguages } = route
+    const { depth, allAvailableLanguages, path } = route
 
     const translatedPath = allAvailableLanguages.get(newLanguage)
 
     if (!translatedPath) { // Route is not translatable
       return {
         status: 'languageNotAvailable',
+        path,
```
Schedulaar: This doesn't make sense here! The `path` doesn't correspond to `newLanguage`!

src/routes/categories/containers/CategoriesContainer.js
klinzo: No because the languages are also not available if they are loading :man_shrugging:

Schedulaar: You have excluded the case above if they are loading.

src/modules/app/StateType.js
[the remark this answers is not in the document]
klinzo: Okay what should we do then?

src/routes/categories/containers/CategoriesContainer.js
klinzo: Right... :man_facepalming:

src/modules/app/StateType.js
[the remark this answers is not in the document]
Schedulaar: Why did you need the change?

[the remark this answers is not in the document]
klinzo: Flow I think, but will have a look again. To display an error refreshProps are needed. (needed in the containers)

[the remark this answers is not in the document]
Schedulaar: Yeah, that's probably the issue. The problem is that we cannot reliably define refreshProps for that special case 🤔
Maybe you can make path optional in the "languageNotAvailable" status, remove the annotated usage in morphcontentlanguage and make refreshprops optional for the error status and only define refreshprops if we have a path in the languageNotAvailable status.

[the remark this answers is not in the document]
klinzo: All right. This mustn't happen anyway, since a language can not be marked as not available if the languages are not available... So I think it does not really matter in any way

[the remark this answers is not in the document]
Schedulaar: I don't really get what you want to say here :see_no_evil:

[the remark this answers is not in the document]
klinzo: The case that the state of the route is `languageNotAvailable` while the state of languages is `error` or `loading` should not be possible. For a language it can only be decided whether it's available if the languages are fetched and `ready`. So since the previously described state should not occur, it does not really matter what we do in the code to satisfy flow.

PR #195, opened 2020-01-27
NATIVE-434: Show back button for language change

klinzo:


src/modules/layout/components/TransparentHeader.js
```diff
@@ -23,19 +23,24 @@ const HorizontalLeft = styled.View`
...
-const BoxShadow: StyledComponent<{}, ThemeType, *> = styled.View`
+const BoxShadow: StyledComponent<{float: boolean}, ThemeType, *> = styled.View`
   background-color: transparent;
-  position: absolute;
-  z-index: 100;
-  top: 0;
-  left: 0;
-  right: 0;
   height: ${props => props.theme.dimensions.modalHeaderHeight};
+  ${props => props.float
```
maxammann: Why is this not needed if a menu is shown? Maybe we should just have one prop `withMenu` and remove the prop `float`. I think the property `float` would require documentation.

klinzo: Imo these two props are not connected, only in this specific case. Am I understanding correctly that you would use only one prop?

src/modules/layout/components/TransparentHeader.js
```diff
@@ -79,17 +84,17 @@ class TransparentHeader extends React.PureComponent<PropsType> {
...
+      <BoxShadow theme={theme} float={float}>
         <Horizontal>
           <HorizontalLeft>
             <HeaderBackButton onPress={this.goBack} />
           </HorizontalLeft>
-          <MaterialHeaderButtons>
+          { withMenu && <MaterialHeaderButtons>
             <Item title={t('share')} show='never' onPress={this.onShare} />
-          </MaterialHeaderButtons>
+          </MaterialHeaderButtons> }
```
Schedulaar: Do we really need the `withMenu` property? We could instead just display the share button if we have `url` in `this.props.navigation.state.params`.
`withMenu` sounds quite abstract for it only deciding to show a `share`-button or not (quite specific).

src/modules/app/createAppContainer.js
```diff
@@ -43,7 +43,10 @@ const createNavigationRouteConfig = (Component: NavigationComponent, header = nu
 })
 
 const transparentHeader = (headerProps: HeaderProps) =>
-  <TransparentHeaderContainer scene={headerProps.scene} scenes={headerProps.scenes} />
+  <TransparentHeaderContainer scene={headerProps.scene} scenes={headerProps.scenes} withMenu={false} float={false} />
+
+const transparentHeaderWithMenu = (headerProps: HeaderProps) =>
+  <TransparentHeaderContainer scene={headerProps.scene} scenes={headerProps.scenes} withMenu float />
```
Schedulaar: I'd rather distinguish between "floating" header and "static" (?) header.

PR #198, opened 2020-01-29
NATIVE-442: Use locale only if available

Schedulaar:


Schedulaar:
It's always English, since we cannot (easily) distinguish German-Simple from any other language...

PR #199, opened 2020-02-01
NATIVE-456: Upgrade sentry

klinzo:


maxammann:
Did you test whether crashes get reported in release mode?

maxammann:
I think we can merge this as long as Hermes is disabled. We can test this before the next release.

Schedulaar:
What does Hermes have to do with it?

maxammann:
Sentry does not yet work with Hermes, I think it is better to test after merging. We have to test sentry nonetheless before a proper release.

klinzo:
I don't care whether to test it now or later. But @maxammann has a point that we have to do it anyway before a release.

klinzo:
@maxammann ping

PR #201, opened 2020-02-10
NATIVE-9: Track webview errors

Schedulaar:


Schedulaar:
Probably someone needs to take a look at the iOS podfiles :thinking:

src/modules/common/components/RemoteContent.js
```diff
@@ -72,18 +76,17 @@ class RemoteContent extends React.Component<PropType, StateType> {
...
+        useWebKit={false}
+        javaScriptEnabled
 
-          dataDetectorTypes={dataDetectorTypes}
-          domStorageEnabled={false}
-          showsVerticalScrollIndicator={false}
+        dataDetectorTypes={dataDetectorTypes}
+        domStorageEnabled={false}
+        showsVerticalScrollIndicator={false}
           showsHorizontalScrollIndicator={false}
```
klinzo: Wrong alignment

src/modules/common/renderHtml.js
```diff
@@ -6,6 +6,36 @@ import type { ThemeType } from '../theme/constants/theme'
...
+      return window.setTimeout(function() { reportError(message) }, 100)
+    }
+
+    window.ReactNativeWebView.postMessage(JSON.stringify({ type: "error", message: message }))
+  };
+  
+  window.onerror = function(msg, url, lineNo, columnNo, error) {
+    // from https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror
+    var string = msg.toLowerCase()
+    var substring = "script error"
```
klinzo: Weird naming, also `string`

Schedulaar: Since this is the referenced code from mozilla, I'd probably leave it like this.

src/modules/common/components/RemoteContent.js
```diff
@@ -46,10 +46,14 @@ class RemoteContent extends React.Component<PropType, StateType> {
     if (!event.nativeEvent) {
       return
     }
-    const height = parseFloat(event.nativeEvent.data)
-    this.setState({
-      webViewHeight: height
-    }, this.props.onLoad)
+    const message = JSON.parse(event.nativeEvent.data)
+    if (message.type === 'error') {
+      throw Error(`An error occurred in the webview:\n${message.message}`)
```
maxammann: The error should also be sent to sentry.

Schedulaar: I cannot find a single instance where we explicitly send an error to sentry. 
Also sentry attaches a global listener to uncaught exceptions and unhandled promise rejections (see https://docs.sentry.io/platforms/javascript/). So we shouldn't need to tell sentry to track the error explicitly... What do you think?

maxammann: You are right, these errors are cought by sentry. This also crashes the app though. Should we only log an error and send it to sentry? (`console.error(); Sentry.log()`)

Schedulaar: No, it doesn't crash the app. since the error is not thrown within a react lifecycle method.

PR #205, opened 2020-02-24
NATIVE-451: Make german the default language

klinzo:


klinzo:
Could not reproduce the issue, for me it's falling back to german, yes.

maxammann:
Oke should be oke then. Lets see if we encounter this again.

PR #206, opened 2020-02-24
NATIVE-430: Search result not clickable when only one result

NKlug:
I have no idea why, but it works. (I gave one of the `View`s a `minHeight`)

### DIFF
--- a/src/routes/search/components/SearchModal.js
+++ b/src/routes/search/components/SearchModal.js
@@ -87,13 +87,16 @@ class SearchModal extends React.Component<PropsType, StateType> {
 
     const filteredCategories = this.findCategories(categories)
     return <ScrollView contentContainerStyle={{ flexGrow: 1 }}>
-      <SpaceBetween>
-        <View><CategoryList categories={filteredCategories} query={query} onItemPress={this.onItemPress}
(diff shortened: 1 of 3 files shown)


src/routes/search/components/SearchModal.js
```diff
@@ -87,13 +87,16 @@ class SearchModal extends React.Component<PropsType, StateType> {
...
     return <ScrollView contentContainerStyle={{ flexGrow: 1 }}>
-      <SpaceBetween>
-        <View><CategoryList categories={filteredCategories} query={query} onItemPress={this.onItemPress}
-                            theme={theme} language={language} /></View>
-        <SearchFeedbackBox t={t} query={query} theme={theme} resultsFound={filteredCategories.length !== 0}
-                           sendFeedback={sendFeedback} />
-      </SpaceBetween>
-    </ScrollView>
+        <SpaceBetween>
+          <View style={{ minHeight: 70 }}>
```
maxammann: This should be calculated dynamically (depending on the device) or based on the theme.

maxammann:
This is ready for merge, right?

maxammann:
Builds should succeed soon again

PR #208, opened 2020-03-02
NATIVE-471: Event body not displayed

Schedulaar:


Schedulaar:
Well, `forceUpdate` already exists in the context of the ContentLoadCriterion and only indicates that JSON stuff should be refreshed. Here I want to say that everything should be refreshed, hence I chose to use another name.

PR #219, opened 2020-03-07
NATIVE-470: Add alt text for header icons

NKlug:


src/modules/layout/components/Header.js
```diff
@@ -160,14 +160,15 @@ class Header extends React.PureComponent<PropsType> {
     return <BoxShadow theme={theme}>
       <Horizontal>
         <HorizontalLeft>
-          {this.canGoBackInStack() ? <HeaderBackButton onPress={this.goBackInStack} /> : <Logo source={logo} />}
+          {this.canGoBackInStack() ? <HeaderBackButton onPress={this.goBackInStack} />
+          : <Logo source={logo} accessibilityLabel={'Integreat'} />}
```
klinzo: Maybe `Home` or `Dashboard` would be better here.

src/modules/layout/components/Header.js
```diff
@@ -144,12 +144,12 @@ class Header extends React.PureComponent<PropsType> {
 
   renderItem (
     title: string, iconName?: string, show: 'never' | 'always',
-    onPress: ?() => void | Promise<void>
+    onPress: ?() => void | Promise<void>, accessibilityLabel: ?string = undefined
```
maxammann: Maybe we should not make this optional. Such that every icon gets an alt text.

src/modules/layout/components/Header.js
klinzo: @maxammann @Schedulaar what do you think?

maxammann: Does this always go to the dashboard?

klinzo: I tink so, yes

Schedulaar: No it's only display when we're already on the Dashboard. So either we use an empty alt text or 'Integreat Logo'. (But probably rather empty)

maxammann: Rather empty those thumb ups are for :)

maxammann:
Merged in develop

PR #232, opened 2020-04-07
NATIVE-443: Fix 2nd level rtl on ios

klinzo:


maxammann:
If you merge in develop then the buids should be fixed.

klinzo:
Partly... Bengin had strange occurances on Android 10 as well, I couldn't reproduce it on my android device.

PR #234, opened 2020-04-09
NATIVE-502: CoronaSnackbar

steffenkleinle:


maxammann (approved):

src/modules/layout/components/Snackbar.js
```diff
@@ -39,9 +47,13 @@ const Action: StyledComponent<{}, ThemeType, *> = styled.Text`
 export type SnackbarActionType = {|
   label: string,
   onPress: () => void | Promise<void>
+|} | {|
+  label: string,
+  href: string
 |}
```
Schedulaar, line 53: It would be better to only have `label` and `onPress` here just like before. You can do the linking by passing an appropriate `onPress` method to Snackbar.

src/modules/layout/containers/PermissionSnackbarContainer.js
```diff
@@ -30,6 +31,7 @@ class PermissionSnackbarContainer extends React.Component<PropsType, StateType>
   constructor (props: PropsType) {
```
Schedulaar, line 31: This would become a `SnackbarContainer` und not a `PermissionSnackbarContainer` anymore.

src/modules/layout/containers/PermissionSnackbarContainer.js
```diff
@@ -98,18 +100,47 @@ class PermissionSnackbarContainer extends React.Component<PropsType, StateType>
...
     const { navigation } = this.props
     const { index, routes } = navigation.state
     const currentRoute = routes[index]
 
-    return currentRoute.key === 'CityContent' && currentRoute.routes[currentRoute.index]?.routeName === 'Dashboard'
+    return this.cityRoute() && currentRoute.routes[currentRoute.index]?.routeName === 'Dashboard'
+  }
+
+  shouldShowCoronaSnackbar = () => {
+    return this.state.showCoronaSnackbar && (this.landingRoute() || this.cityRoute())
```
Schedulaar, line 120: I thought we only want to show it in the cityRoute?

michael-markl (approved):
We should wait with merging until we get feedback from Google. If no change on our side is required I'd vote for not-merging this at all.

src/modules/layout/containers/PermissionSnackbarContainer.js
klinzo: Not that I know of. At least not if it is required by google. Because if it is required by google, it should be shown as soon as possible. Or did I miss something here?

klinzo:
Yeah I'll definitely wait with merging, but let us discuss this next week.

klinzo:
Closed. See https://integreat.slack.com/archives/CDV9QF1GA/p1586854815001900 for more information.

PR #235, opened 2020-04-09
NATIVE-419: Fixed error handling when API requests are not possible. …

lulo4907:
…Added async and await expression to complete sendFeedback promises.

### DIFF
--- a/src/routes/feedback/components/FeedbackModal.js
+++ b/src/routes/feedback/components/FeedbackModal.js
@@ -54,7 +54,7 @@ class FeedbackModal extends React.Component<PropsType, StateType> {
 
   onFeedbackCommentChanged = (comment: string) => this.setState({ comment })
 
-  onSubmit = () => {
+  onSubmit = async () => {
     const { feedbackIndex, comment } = this.state
     const feedbackItem = this.props.feedbackItems[feedbackIndex]
     const feedbackData: FeedbackParamsType = {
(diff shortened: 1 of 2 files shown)


src/routes/feedback/components/FeedbackModal.js
```diff
@@ -54,7 +54,7 @@ class FeedbackModal extends React.Component<PropsType, StateType> {
 
   onFeedbackCommentChanged = (comment: string) => this.setState({ comment })
 
-  onSubmit = () => {
+  onSubmit = async () => {
```
Schedulaar, line 57: This `async` is not needed here, as the promises are handled using `then` and `catch` below.
```suggestion
  onSubmit = () => {
```

src/routes/feedback/containers/FeedbackModalContainer.js
```diff
@@ -18,7 +18,7 @@ class FeedbackModalContainer extends React.Component<{| navigation: NavigationSc
   sendFeedback = async (feedbackData: FeedbackParamsType) => {
     const apiUrl = await determineApiUrl()
     const feedbackEndpoint = createFeedbackEndpoint(apiUrl)
-    feedbackEndpoint.request(feedbackData)
+    await feedbackEndpoint.request(feedbackData)
```
Schedulaar, line 21: :+1: Good catch.

Schedulaar:
@klinzo In fact, when I implemented the feedback I already implemented displaying an error message when sending failed, but it seems I have forgot to check if it works. Turns out there was an `await` missing that @lulo4907 added.

PR #241, opened 2020-04-16
NATIVE-499: Failed to delete resource cache

klinzo:


klinzo:
@maxammann yes, maybe I should add a comment. If you are offline and try to open a city, it gets saved that you opened the city (which is fine I'd say), but since no content or resource data is fetched, nothing is saved and therefore folders are not created.

maxammann:
In that case the resource cache object probably shouldn't be filled with paths to files which do not exist, right?

I think we should try to keep the object and the files on disk in sync.

klinzo:
No, it is not filled with paths that do not exist, at least not that I can see. The only thing happens is: If a city is opened, an entry about this usage incl. timestamp is saved, no matter if an internet connection is available or not. I think this is perfectly fine and is even the right way to go.

When deleting old files, we just go through all last usages (map of city code and timestamp) and try to delete the data of all cities except the latest 3 cities. So in fact, no paths and also no not existing paths are saved.

maxammann:
k, but then the files in the resource cache we are trying to delete here do not exist sometimes.

I think this needs further investigation.

klinzo:
No, it does not need further investigating, everything is clear here. :D I'll try to explain again. Upon deleting the files of older cities, we don't have a map of paths we have to clear. Instead, we have a map of last usages of cities. Now, we just go through this map and try to delete the files and dirs of all cities which usages are not on of the latest three usages.

If we don't have an internet connection and try to open a city, e.g. `augsburg`, a last usage timestamp for that opened city is saved, but no files are saved for the city since nothing is fetched without an internet connection. So, upon deleting old files we just iterate through all last usages. We find an entry `augsburg` and try to delete the resource cache and the files for said city `augsburg`. These files are not existing though. Anyway. imo this is perfectly fine. If you still have any questions, I'm on mumble.

PR #242, opened 2020-04-16
NATIVE-501: Failed to write manifest file

klinzo:
This aims to resolve this sentry issue: https://sentry.integreat-app.de/tur-an-tur-digitalfabrik/integreat-react-native-app/issues/90/?query=is%3Aunresolved

As mentioned here https://github.com/react-native-community/async-storage/issues/88, this issue may be fixed with upgrading async-storage. I did not have a clue on how to reproduce it, so  I'd just try upgrading it as done in this PR.

The upgrade to v1.8 includes breaking changes (as mentioned here: https://github.com/react-native-community/async-storage/releases/tag/v1.8.0). However, no other library is using async-storage, so it should be fine. Also, I tried migrating and using an existing app installation which worked without problems. Any other opinions?

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -240,7 +240,7 @@ PODS:
     - React
   - rn-fetch-blob (0.12.0):
     - React-Core
-  - RNCAsyncStorage (1.8.1):
+  - RNCAsyncStorage (1.9.0):
     - React
   - RNFastImage (7.0.2):
     - React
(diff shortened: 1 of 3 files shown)


maxammann:
Seems to run on iOS :+1:

klinzo:
@maxammann should I change this so it does not run on ios to get your approval? :stuck_out_tongue:

maxammann:
Needs to run now on windows mobile aswell :man_shrugging:

PR #243, opened 2020-04-20
NATIVE-516: Add design proposals

klinzo:


klinzo:
This were all proposals I found. If you know about anything else, please let me know.

maxammann:
Should we move this do `docs`? There is already a documentation file about ui/ux changes. Maybe we can describe them there and link.

klinzo:
Not sure, I also thought about it. I'd say linking (or at least mentioning) it in the ui-ux doc would definitely make sense, not sure about moving everything there since these are not really docs. So I'd vote for mentioning/linking but not moving.

klinzo:
@lulo4907 what do you mean by that? Where did you get that name/information from that it should be here?

lulo4907:
@klinzo Ah, okay. Then I misunderstood some content of this task.

PR #246, opened 2020-04-29
NATIVE-437: Rename StatusBar

klinzo:
I could not reproduce the bug described in the issue: https://issues.integreat-app.de/browse/NATIVE-437. Probably this was fixed as described in this comment https://github.com/facebook/react-native/issues/26619#issuecomment-541143915  with the upgrade on ReactNative 0.61.5. So I just renamed `AndroidStatusBar` to `StatusBar` since it's not android exclusive (see: https://reactnative.dev/docs/statusbar). Don't know why git didn't get that.

If the behaviour is different for you and you can reproduce it let me know.


maxammann:
I checked on an iphone with latest version. The bug is gone :)

src/modules/app/components/StatusBar.js
```diff
@@ -0,0 +1,19 @@
...
+import { StatusBar as ReactNativeStatusBar } from 'react-native'
+
+type PropsType = {|
+  theme: ThemeType
+|}
+
+class StatusBar extends React.PureComponent<PropsType> {
+  render () {
+    return (
+      <ReactNativeStatusBar backgroundColor={this.props.theme.colors.backgroundAccentColor} barStyle='dark-content' />
```
maxammann, line 14: I don' really remember what barStyle does here :/ 
But actually I think it makes sense that the text it always is black.

klinzo: It does exactly what the goal of the Issue was. Notify Android/iOS that the text is dark, and therefore the status bar has to be white.
From the docs `barStyle

Sets the color of the status bar text.`

lulo4907: I've found options for theming the app (`brightColors`, `darkColors`; theme.js) but had no access at changing it with the JS files. What I think is, that if `darkColors` is a real option for choosing at the CMS (I quite have no access to the CMS at the moment), the static value `dark-content` of barStyle would lead into a misconfiguration.

klinzo: @lulo4907 I don't really understand your point. :thinking: All colors are defined on our side, we don't get anything from the cms.

maxammann: @lulo4907 is right about the theming problem, right now the dark theme is not used. Maybe we should remove it for now until we want to add a dark mode.

klinzo: As long it is not used I don't see a problem? At least I don't really see a point in removing it now.

klinzo: We just have to keep an eye on this upon implementing a dark mode.

klinzo: Created an issue with a note to keep this in mind: https://issues.integreat-app.de/browse/NATIVE-526

PR #248, opened 2020-05-05
NATIVE-524: Hide news and extras if disabled

Schedulaar:


src/modules/endpoint/sagas/loadEvents.js
```diff
@@ -9,6 +9,7 @@ import determineApiUrl from '../determineApiUrl'
 function * loadEvents (
   city: string,
   language: string,
+  eventsEnabled: boolean,
```
klinzo, line 12: Maybe we should switch to an object as param at some point.

src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -81,7 +83,7 @@ export default function * loadCityContent (
 
   const [categoriesMap, events] = yield all([
     call(loadCategories, newCity, newLanguage, dataContainer, shouldUpdate),
-    call(loadEvents, newCity, newLanguage, dataContainer, shouldUpdate)
+    call(loadEvents, newCity, newLanguage, cityModel.eventsEnabled, dataContainer, shouldUpdate)
```
klinzo, line 86: Why do we have to set events to an empty array if events are disabled? Can't we just return an empty array here if disabled?

Schedulaar: The data container has to be set as well, since other sagas might as well query events from the data container (and should receive an empty array).

klinzo: Might query or are querying?

Schedulaar: Are querying

src/modules/endpoint/sagas/loadEvents.js
Schedulaar: Yeah, but I don't want to bloat this PR.

Schedulaar:
As these are changes to the core sagas, I'd also explicitly request a review from @maxammann.

src/modules/endpoint/sagas/__mocks__/loadCities.js
```diff
@@ -11,7 +11,7 @@ function * loadCities (
 ): Saga<Array<CityModel>> {
   const citiesAvailable = yield call(() => dataContainer.citiesAvailable())
 
-  if (!citiesAvailable || forceRefresh) {
+  if (!citiesAvailable) {
```
maxammann, line 14: Why was this necessary? this should behave in the same way as the non-mock loadCities.js.

`forceRefresh` should probably not be true in the tests.

maxammann: This is because the tests should not depend on refreshing data and calling the DataContainer.

Schedulaar: Because it is necessary to also test, that if the lastUpdate is quite a while ago, `loadCityContent` would call loadCities with `forceRefresh===true`.
(Just like `loadCategories` and  `loadEvents` and `loadLanguages` btw.)

PR #250, opened 2020-05-06
NATIVE-527: Fix signing if the keystore is not configured

Schedulaar:


klinzo:
@maxammann so should we merge this? I thought you said this is kind of redundant/outdated?

maxammann:
I think we should provide a way of building the app without `fastlane` as it can be difficult to setup.
Let's leave this open until I finished the docu. Maybe there is a more elegant way.

maxammann:
I adressed this in #255

PR #252, opened 2020-05-06
NATIVE-529: Set release times properly

maxammann:


klinzo:
@maxammann you should probably rebase to develop, I had the same problem.

.circleci/config.yml
```diff
@@ -47,7 +47,7 @@ workflows:
   weekly:
     triggers:
       - schedule:
-          cron: "0 0 18,20,24 * *" # FIXME "0 0 * * 3"
+          cron: "0 3 * * 4"
```
klinzo, line 50: The syntax is quite strange imo, should we add an inline comment (same below)?

maxammann: I'll add a link to https://crontab.guru/

PR #253, opened 2020-05-11
NATIVE-530: Add design guide and color scheme

klinzo:


docs/13-ui-ux.md
```diff
@@ -7,4 +7,7 @@ All new design proposals and mockups should be added to the ui-ux folder AND the
 Access to the drive can be requested at one of the members of Integreat mentioned in the [README](../README.md).
 
 ## Icons
-Current icons can be found [here in the drive](https://drive.google.com/drive/folders/1FlSvpC0et0OnFuV-rxV1m2U3hoZO-ufJ).
+The current icons can be found [here in the drive](https://drive.google.com/drive/folders/1FlSvpC0et0OnFuV-rxV1m2U3hoZO-ufJ).
+
+## Color scheme
+The current color schmeme can be found [here](../ui-ux/2017_11_ColorScheme/ColorScheme.png) or in the [drive](https://drive.google.com/drive/folders/19fv7ot0wO8npxHWMQCzGsuCe6Lzndzny).
```
Schedulaar, line 13: Should we have a duplication here? I think, it would be enough to have the scheme in the drive and put a link here?

klinzo: Maybe we shouldn't add files to the repo here at all.. all resources are duplicated right now... Not sure about the best approach here.

Schedulaar: Imo it is usefull and sufficient if we had a link to the corresponding Design and ColorScheme folders on GoogleDrive in this repo.

klinzo: I think so too...all right I guess I will remove everything again then... :D

klinzo: @maxammann any objections?

maxammann: I think we should decide which information should be in the drive and which in the repo. Drive is for internal (and secret) stuff. Everything else should live in a repository as we are open-source.

Duplication is bad ofc, as we kind of manage the space where the files live in the drive we should remove them there. We should also have a workflow for design stuff in the future. Work which is related to the app should be checked in somehow imo.

@klinzo please squash commits in this branch if this is merged. Else the PDFs still take space after merging.

klinzo: So you would keep it in the drive and the repo? Should we discuss this in the call tomorrow?

maxammann:
Should we put this as draft until we decided?

PR #254, opened 2020-05-19
NATIVE-535: Use white background for feedback icons

klinzo:


klinzo:
In contrast to the issue description we decided to just use a white background for the feedback icons, no matter whether for integreat nor for malte. The discussion can be found here and in the previous messages: https://integreat.slack.com/archives/CKDL54J1H/p1589888015003000

klinzo:
image

PR #256, opened 2020-05-24
NATIVE-482: Add missing accessibility roles and hints

NKlug:
I tested all the clickables with a screenreader again (only on Android though) and added a few missing hints and roles. IMO everything works pretty good already, I didn't have to change much.
I'm not entirely sure if I merged the locales correctly, so please check.


klinzo:
No other accessibility roles/hints needed? How did you proceed to decided where roles are needed and where not?

NKlug:
The task says to not over-do it with the accessibility roles and hints. In order to be consistent I decided not to add additional hints to buttons etc., which leaves only a few elements where I figured it might be helpful (e.g. a link should not be treated as a button). Apart from that, Android Talkback already does quite a good job with most elements.

PR #257, opened 2020-05-31
WEBAPP-518: Fix api client flow types

Schedulaar:

### DIFF
--- a/src/modules/app/__tests__/CategoriesRouteStateView.spec.js
+++ b/src/modules/app/__tests__/CategoriesRouteStateView.spec.js
@@ -14,7 +14,9 @@ describe('CategoriesRouteStateView', () => {
       parentPath: '',
       order: 0,
       availableLanguages: new Map(),
-      lastUpdate: moment('2017-11-18 19:30:00', moment.ISO_8601)
+      lastUpdate: moment('2017-11-18 19:30:00', moment.ISO_8601),
+      thumbnail: '/thumbnail.jpg',
+      hash: '123asdf'
     }),
     '/augsburg/de/erste-hilfe': new CategoryModel({
       root: false,
@@ -24,7 +26,9 @@ describe('CategoriesRouteStateView', () => {
       parentPath: '/augsburg/de',
       order: 0,
       availableLanguages: new Map(),
-      lastUpdate: moment('2017-11-18 19:30:00', moment.ISO_8601)
+      lastUpdate: moment('2017-11-18 19:30:00', moment.ISO_8601),
+      thumbnail: '/thumbnail.jpg',
+      hash: '123asdf'
     }),
     '/augsburg/de/zweite-hilfe': new CategoryModel({
       root: false,
@@ -34,7 +38,9 @@ describe('CategoriesRouteStateView', () => {
       parentPath: '/augsburg/de/erste-hilfe',
       order: 0,
       availableLanguages: new Map(),
-      lastUpdate: moment('2017-11-18 19:30:00', moment.ISO_8601)
+      lastUpdate: moment('2017-11-18 19:30:00', moment.ISO_8601),
+      thumbnail: '/thumbnail.jpg',
+      hash: '123asdf'
     })
   }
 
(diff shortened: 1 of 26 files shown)


src/modules/endpoint/DatabaseConnector.js
```diff
@@ -372,6 +374,8 @@ class DatabaseConnector {
         live: jsonObject.live,
         eventsEnabled: jsonObject.events_enabled,
         extrasEnabled: jsonObject.extras_enabled,
+        tunewsEnabled: false, // todo: NATIVE-549
```
klinzo, line 377: Strange TODOs :thinking:

src/modules/endpoint/ResourceURLFinder.js
```diff
@@ -5,9 +5,14 @@ import { Parser } from 'htmlparser2'
 import type { FetchMapType } from './sagas/fetchResourceCache'
 import { reduce } from 'lodash'
 import hashUrl from './hashUrl'
-import { ExtendedPageModel } from '@integreat-app/integreat-api-client'
 import Url from 'url-parse'
 
+interface InputEntryType {
```
klinzo, line 10: Why use `interface` here? What is the difference?

src/routes/events/components/Events.js
```diff
@@ -121,8 +122,8 @@ class Events extends React.Component<PropsType> {
           <>
             <PageDetail identifier={t('date')} information={event.date.toFormattedString(language)}
                         theme={theme} language={language} />
-            {event.location.location && <PageDetail identifier={t('location')}
-                                                    information={event.location.location} theme={theme}
+            {location && <PageDetail identifier={t('location')}
+                                                    information={location} theme={theme}
```
klinzo, line 126: Indentation.

src/modules/endpoint/ResourceURLFinder.js
Schedulaar: https://flow.org/en/docs/types/interfaces/
The benefit is, that I can test the method by only passing objects that have the necessary properties and so I don't have to create ExtendedPageModels with all the attributes that are not needed for (testing) this method.

src/modules/endpoint/DatabaseConnector.js
Schedulaar: Well, these attributes have to be properly and carefully filled into the database. I didn't want to do this in this task.

src/modules/endpoint/ResourceURLFinder.js
klinzo: Should we mention this in the model builder issue then to change it?

src/modules/endpoint/DatabaseConnector.js
klinzo: Was more like `todo` vs `TODO` but oc that's fine :smile:

src/modules/endpoint/ResourceURLFinder.js
Schedulaar: Not sure what you mean by that

src/modules/app/__tests__/CategoriesRouteStateView.spec.js
```diff
@@ -14,7 +14,9 @@ describe('CategoriesRouteStateView', () => {
       parentPath: '',
       order: 0,
       availableLanguages: new Map(),
-      lastUpdate: moment('2017-11-18 19:30:00', moment.ISO_8601)
+      lastUpdate: moment('2017-11-18 19:30:00', moment.ISO_8601),
```
maxammann, line 17: Use there the syntax: `2007-08-31T16:47+00:00`

src/routes/dashboard/containers/__tests__/DashboardContainer.spec.js
```diff
@@ -82,6 +83,9 @@ describe('DashboardContainer', () => {
   }
 
   const rootCategory = categoriesMap.findCategoryByPath(`/${city.code}/${language.code}`)
+  if (!rootCategory) {
+    throw Error('The root category was not found. Sth\'s odd.')
```
maxammann, line 87:
```suggestion
    throw Error('The root category was not found. Something is very odd.')
```

src/modules/endpoint/ResourceURLFinder.js
klinzo: This (the use of interface) will be obsolete once we have model builders, right? If so, we could add a note to the model builder issue you created.

Schedulaar: Hmmm, no I wouldn't adjust this. Imo this is equally good or even better than using modelBuilders and ExtendedPageModel here, as we only need the 3 attributes defined in the interface for using the method...

PR #263, opened 2020-06-08
NATIVE-554: Make alias search less obstrusive

SimonVortkamp:

I need to fix 2 lines which are too long at the moment. (lint max-len)

### DIFF
--- a/src/routes/landing/components/CityEntry.js
+++ b/src/routes/landing/components/CityEntry.js
@@ -62,13 +62,13 @@ class CityEntry extends React.PureComponent<PropType> {
                     underlayColor={theme.colors.backgroundAccentColor}>
         <View>
           <Label theme={theme} searchWords={[filterText]} textToHighlight={city.name} sanitize={normalizeSearchString}
-                 highlightStyle={{ backgroundColor: theme.colors.themeColor }} />
+                 highlightStyle={{ backgroundColor: theme.colors.backgroundColor, fontWeight: 'bold' }} />
           {aliases.length > 0 && <Aliases>
             {aliases.map(
               (alias, index) => <AliasLabel key={alias} theme={theme} searchWords={[filterText]}
                                             textToHighlight={index === aliases.length - 1 ? alias : `${alias}, `}
                                             sanitize={normalizeSearchString}
-                                            highlightStyle={{ backgroundColor: theme.colors.themeColor }} />
+                                            highlightStyle={{ backgroundColor: theme.colors.backgroundColor, fontWeight: 'bold' }} />
             )}
           </Aliases>}
         </View>
--- a/src/routes/landing/components/CityEntry.js
+++ b/src/routes/landing/components/CityEntry.js
@@ -10,6 +10,8 @@ import normalizeSearchString from '../../../modules/common/normalizeSearchString
 import { View } from 'react-native'
 import Highlighter from 'react-native-highlight-words'
 
+const MAX_NUMBER_OF_ALIASES_SHOWN = 3
+
 const CityListItem: StyledComponent<{}, {}, *> = styled.TouchableHighlight`
   padding: 7px;
   width: 100%;
@@ -64,9 +66,9 @@ class CityEntry extends React.PureComponent<PropType> {
           <Label theme={theme} searchWords={[filterText]} textToHighlight={city.name} sanitize={normalizeSearchString}
                  highlightStyle={{ backgroundColor: theme.colors.backgroundColor, fontWeight: 'bold' }} />
           {aliases.length > 0 && <Aliases>
-            {aliases.map(
+            {aliases.slice(0, MAX_NUMBER_OF_ALIASES_SHOWN).map(
               (alias, index) => <AliasLabel key={alias} theme={theme} searchWords={[filterText]}
-                                            textToHighlight={index === aliases.length - 1 ? alias : `${alias}, `}
+                                            textToHighlight={index !== aliases.slice(0, MAX_NUMBER_OF_ALIASES_SHOWN).length - 1 ? `${alias}, ` : aliases.length <= MAX_NUMBER_OF_ALIASES_SHOWN ? alias : `${alias},...`}
                                             sanitize={normalizeSearchString}
                                             highlightStyle={{ backgroundColor: theme.colors.backgroundColor, fontWeight: 'bold' }} />
             )}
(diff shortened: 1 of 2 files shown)


src/routes/landing/components/CityEntry.js
```diff
@@ -62,13 +64,13 @@ class CityEntry extends React.PureComponent<PropType> {
...
         <View>
           <Label theme={theme} searchWords={[filterText]} textToHighlight={city.name} sanitize={normalizeSearchString}
-                 highlightStyle={{ backgroundColor: theme.colors.themeColor }} />
+                 highlightStyle={{ backgroundColor: theme.colors.backgroundColor, fontWeight: 'bold' }} />
           {aliases.length > 0 && <Aliases>
-            {aliases.map(
+            {aliases.slice(0, MAX_NUMBER_OF_ALIASES_SHOWN).map(
               (alias, index) => <AliasLabel key={alias} theme={theme} searchWords={[filterText]}
-                                            textToHighlight={index === aliases.length - 1 ? alias : `${alias}, `}
+                                            textToHighlight={index !== aliases.slice(0, MAX_NUMBER_OF_ALIASES_SHOWN).length - 1 ? `${alias}, ` : aliases.length <= MAX_NUMBER_OF_ALIASES_SHOWN ? alias : `${alias},...`}
```
klinzo, line 71: Line too long and unclear. Also, afaics, the comma and dots (`,...`) are also part of an alias, that should not be the case. Also, a space is missing between the comma and the dots and this way it won't work for RTL layouts (as it is always `, ...` and never `... ,` (same for aliases) as it should be for RTL layouts. Therefore, you'd need to split `alias`, `,` and `...`. Maybe also extract some logic here as it gets quite unclear.

src/routes/landing/components/CityEntry.js
```diff
@@ -62,13 +64,13 @@ class CityEntry extends React.PureComponent<PropType> {
                     underlayColor={theme.colors.backgroundAccentColor}>
         <View>
           <Label theme={theme} searchWords={[filterText]} textToHighlight={city.name} sanitize={normalizeSearchString}
-                 highlightStyle={{ backgroundColor: theme.colors.themeColor }} />
+                 highlightStyle={{ backgroundColor: theme.colors.backgroundColor, fontWeight: 'bold' }} />
```
klinzo, line 67: I think `backgroundColor: theme.colors.backgroundColor` is not needed at all, is it? Same for other PRs and the occurrence below.

SimonVortkamp: You're right. In the webapp it was neccesary, that's why I assumed it works the same way in the native app, but I was wrong

klinzo: Do you know why it is necessary in the webapp and not in the native app?

src/routes/landing/components/CityEntry.js
```diff
@@ -60,18 +74,28 @@ class CityEntry extends React.PureComponent<PropType> {
...
-                                            textToHighlight={index === aliases.length - 1 ? alias : `${alias}, `}
-                                            sanitize={normalizeSearchString}
-                                            highlightStyle={{ backgroundColor: theme.colors.themeColor }} />
+            {aliases.slice(0, MAX_NUMBER_OF_ALIASES_SHOWN).map(
+              (alias, index) => <>
+                <AliasLabel key={alias} theme={theme} searchWords={[filterText]}
+                            textToHighlight={alias}
+                            sanitize={normalizeSearchString}
+                            highlightStyle={{ fontWeight: 'bold' }} />
+                {index !== aliases.slice(0, MAX_NUMBER_OF_ALIASES_SHOWN).length - 1 && <>
```
maxammann, line 87: I think it would be good style to avoid the duplicate `aliases.slice(0, MAX_NUMBER_OF_ALIASES_SHOWN)`. This could be achieved by overwriting aliases and having a boolean:
```
slice_needed = aliases.length > MAX_NUMBER_OF_ALIASES_SHOWN
aliases = aliases.slice(0, MAX_NUMBER_OF_ALIASES_SHOWN)
```

src/routes/landing/components/CityEntry.js
```diff
@@ -60,18 +74,28 @@ class CityEntry extends React.PureComponent<PropType> {
...
+                  <Separator theme={theme}>,</Separator>
+                  <Separator theme={theme}> </Separator></>}
+              </>
             )}
+            {aliases.length > MAX_NUMBER_OF_ALIASES_SHOWN && <>
```
maxammann, line 92:
```suggestion
            {slice_needed && <>
```

src/routes/landing/components/CityEntry.js
SimonVortkamp: In the WebApp we use a different highlighter, which has a default color, I think

src/routes/landing/components/CityEntry.js
```diff
@@ -56,22 +70,34 @@ class CityEntry extends React.PureComponent<PropType> {
   render () {
     const { city, theme, filterText } = this.props
     const normalizedFilter = normalizeSearchString(filterText)
-    const aliases = this.getMatchingAliases(city, normalizedFilter)
+    let aliases = this.getMatchingAliases(city, normalizedFilter)
+    const sliceNeeded = aliases.length > MAX_NUMBER_OF_ALIASES_SHOWN
+    aliases = sliceNeeded ? aliases.slice(0, MAX_NUMBER_OF_ALIASES_SHOWN) : aliases
```
klinzo, line 75: From the docs of Array.prototype.slice(): `If end is greater than the length of the sequence, slice extracts through to the end of the sequence (arr.length).` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice)

Therefore the following is enough:
```suggestion
    const aliases = this.getMatchingAliases(city, normalizedFilter).slice(0, MAX_NUMBER_OF_ALIASES_SHOWN)
    const sliceNeeded = aliases.length > MAX_NUMBER_OF_ALIASES_SHOWN
```

src/routes/landing/components/CityGroup.js
```diff
@@ -28,8 +30,8 @@ type PropsType = {|
 class CityGroup extends React.Component<PropsType> {
   render () {
     return <BorderWrapper theme={this.props.theme}>
-      <GroupText theme={this.props.theme}>{this.props.children}</GroupText>
-    </BorderWrapper>
+        <GroupText theme={this.props.theme}>{this.props.children}</GroupText>
```
maxammann, line 33: Indentation is a little bit off here :P

PR #265, opened 2020-06-12
NATIVE-565: Remove glossary and link to new one

maxammann:


maxammann:
@klinzo you are right as soon as Extra is gone in our team we can remove it. Or we keep it in some legacy section.

PR #276, opened 2020-07-08
NATIVE-580: Only deliver every two weeks to beta

klinzo:


.circleci/src/@workflows.yml
```diff
@@ -60,10 +60,10 @@ workflows:
           requires:
             - bump_version
 
-  weekly_development_delivery:
+  bi_weekly_development_delivery:
     triggers:
       - schedule:
-          cron: "0 3 * * 4" # For sanity checks for this syntax: https://crontab.guru/
+          cron: "0 3 8,22 * *" # For sanity checks for this syntax: https://crontab.guru/
```
Taggotty, line 66: Why is this value different from the one above?

Taggotty: And why do we need to define 2 schdules? One in the workflow and one in the task itself?

klinzo: What value above do you mean? The `config.yml` is put together by the other config files, e.g. this `@workflow.yml` file.

There are two schedules: Development delivery and production delivery.

klinzo:
@Taggotty anything to add or can I merge this?

PR #280, opened 2020-07-22
NATIVE-578: Split circle ci config

klinzo:


klinzo:
> Looks great -- but let me know, how is the "src" to the outsourced files path attached?

@lulo4907 sorry I'm not sure if I'm understanding your question correctly? Do you mean how the config in src is transformed to the combined `config.yml`? It uses the circleci command line tools: https://circleci.com/docs/2.0/local-cli/#packing-a-config

Have a look at the package.json and the command `circleci:update-config` for the actual command we are using now.

PR #281, opened 2020-07-24
NATIVE-557: Build configs for gradle

klinzo:


android/app/buildConfigs.gradle
```diff
@@ -0,0 +1,52 @@
...
+}
+
+createConfig()
+
+android {
+    defaultConfig {
+        project.buildConfig.each { k, v ->
+            def escaped = v.replaceAll("%","\\\\u0025")
+            // Make build config values available as string resource, e.g. for use in AndroidManifest
+            resValue "string", k, "\"$escaped\""
```
maxammann, line 49: :+1:

android/app/buildConfigs.gradle
```diff
@@ -0,0 +1,52 @@
...
+        return
+    }
+    println("Created xcconfg file: $proc.in.text")
+
+
+    File f = new File("$project.rootDir/buildConfig.tmp.xcconfig");
+
+    f.eachLine { line ->
+        def parts = line.split(" = ")
+        buildConfig.put(parts[0], parts[1])
```
maxammann, line 35: Maybe some error checking would be nice here :)

maxammann: We also could use here a Java Properties file to load the xcconfig file: https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html

android/app/build.gradle
```diff
@@ -128,7 +129,8 @@ android {
     }
 
     defaultConfig {
-        applicationId "tuerantuer.app.integreat"
+        // TODO Add application id to build configs
+        applicationId project.buildConfig.getOrDefault("BUILD_CONFIG_APPLICTION_ID", "tuerantuer.app.integreat")
```
maxammann, line 133: I think you can remove this and add it to buildConfigs.gradle's android block

android/app/buildConfigs.gradle
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    println("Build config: $buildConfig")
+    project.ext.set("buildConfig", buildConfig)
+}
+
+createConfig()
+
+android {
+    defaultConfig {
```
maxammann, line 45: applicationId could be added here

android/fastlane/Fastfile
```diff
@@ -37,8 +37,10 @@ desc "Create an Android build in release mode. Set the environment variable E2E_
 lane :build do |options|
   version_code = options[:version_code]
   version_name = options[:version_name]
+  build_config_name = options[:build_config_name]
 
   ENV["BUNDLE_CONFIG"] = "./metro.config.ci.js"
+  ENV["BUILD_CONFIG_NAME"] = build_config_name
```
maxammann, line 43: I'm not sure but maybe it makes sense not to pass this as environment variable but as Gradle Property. See below.

maxammann:
```
properties: {
          :VERSION_CODE => version_code,
          :BUILD_CONFIG_NAME => build_config_name,
          :KEYSTORE_PATH => KEYSTORE_PATH,
      }
```

klinzo: Okay, yes, I like the idea and didn't think of it. However, to make it possible to also specify the build config from a script in the package.json, this needs to happen via environment variables. What do you think?

klinzo: Discussed: Environment variables are necessary for the scripts for now.

android/app/buildConfigs.gradle
klinzo: Good suggestion :+1:

android/app/buildConfigs.gradle
```diff
@@ -0,0 +1,55 @@
...
+    proc.waitFor()
+
+    if (proc.exitValue() != 0) {
+        println("Failed to create xcconfig file: $proc.err.text")
+        return
+    }
+    println("Created xcconfg file: $proc.in.text")
+
+    def buildConfigProperties = new Properties()
+    buildConfigProperties.load(new FileInputStream("$project.rootDir/buildConfig.tmp.xcconfig"))
```
maxammann, line 31: The FileInputStream should be closed afterwards, also maybe exceptions need to be handled

klinzo: I don't see a point in handling exceptions right now, shouldn't happen because it is perfect code :stuck_out_tongue: . Also, if an exception occurs, we have to investigate why it fails anyway. Closing the input stream though.

.gitignore
```diff
@@ -35,8 +35,9 @@ yarn-error.log*
 # CocoaPods
 /ios/Pods/
 
-# Temp buildConfig file
+# Temp buildConfig files
```
Schedulaar, line 38: :+1:

PR #285, opened 2020-07-29
NATIVE-586: Run gradle clean before building

klinzo:


.circleci/config.yml
```diff
@@ -108,7 +108,7 @@ jobs:
                 condition: << parameters.build_for_delivery >>
                 steps:
                     - run:
-                        command: bundle exec fastlane build version_name:${NEW_VERSION_NAME} version_code:${NEW_VERSION_CODE}
+                        command: ./gradlew clean | bundle exec fastlane build version_name:${NEW_VERSION_NAME} version_code:${NEW_VERSION_CODE}
```
maxammann, line 111: I think this should be `&`.

maxammann: Else the second command runs even if the first one fails

.circleci/config.yml
```diff
@@ -108,7 +108,7 @@ jobs:
                 condition: << parameters.build_for_delivery >>
                 steps:
                     - run:
-                        command: ./gradlew clean | bundle exec fastlane build version_name:${NEW_VERSION_NAME} version_code:${NEW_VERSION_CODE}
+                        command: ./gradlew clean & bundle exec fastlane build version_name:${NEW_VERSION_NAME} version_code:${NEW_VERSION_CODE}
```
maxammann, line 111: Probably should be `&&`

klinzo:
```suggestion
                        command: ./gradlew clean && bundle exec fastlane build version_name:${NEW_VERSION_NAME} version_code:${NEW_VERSION_CODE}
```

klinzo: Yep, forgot to update the config.

PR #286, opened 2020-08-03
NATIVE-504: Move release notes and create jira releases

klinzo:
This pull request belongs to an issue on our [bugtracker](https://issues.integreat-app.de/).
You can find it there by looking for an issue with the key which is mentioned in the title of this pull request.
It starts with the keyword **NATIVE**.

PR #292, opened 2020-08-11
NATIVE-587: Use release notes

klinzo:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -444,19 +444,16 @@ jobs:
                     - ~/.cache/yarn
                     - node_modules
             - run:
-                command: export GITHUB_RELEASE_NOTES=$(tools/parse-release-notes parse --ios --android --development)
-                name: Parse github release notes
-            - run:
-                command: tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/de-DE/changelogs/default.txt --language de
+                command: tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/de-DE/changelogs/default.txt --language de --production
                 name: Parse german android release notes
             - run:
-                command: tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/en-GB/changelogs/default.txt --language en && tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/en-US/changelogs/default.txt --language en
+                command: tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/en-GB/changelogs/default.txt --language en --production && tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/en-US/changelogs/default.txt --language en --production
                 name: Parse english android release notes
             - run:
-                command: tools/parse-release-notes parse --ios --destination ios/fastlane/metadata/de-DE/release_notes.txt --language de
+                command: tools/parse-release-notes parse --ios --destination ios/fastlane/metadata/de-DE/release_notes.txt --language de --production
                 name: Parse german ios release notes
             - run:
-                command: tools/parse-release-notes parse --ios --destination ios/fastlane/metadata/en-US/release_notes.txt --language en
+                command: tools/parse-release-notes parse --ios --destination ios/fastlane/metadata/en-US/release_notes.txt --language en --production
                 name: Parse english ios release notes
             - when:
                 condition: << parameters.production_delivery >>
@@ -468,7 +465,7 @@ jobs:
                         command: tools/move-release-notes move-to ${NEW_VERSION_NAME} --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} --branch ${CIRCLE_BRANCH} --commit-sha ${CIRCLE_SHA1}
                         name: Move release notes
             - run:
-                command: tools/bump-version bump-to ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} --branch ${CIRCLE_BRANCH} --release-notes ${GITHUB_RELEASE_NOTES} <<# parameters.production_delivery >> --development-release <</ parameters.production_delivery >>
+                command: tools/bump-version bump-to ${NEW_VERSION_NAME} ${NEW_VERSION_CODE} --deliverino-private-key ${DELIVERINO_PRIVATE_KEY} --owner ${CIRCLE_PROJECT_USERNAME} --repo ${CIRCLE_PROJECT_REPONAME} --branch ${CIRCLE_BRANCH} --release-notes "${tools/parse-release-notes parse --ios --android}" <<# parameters.production_delivery >> --development-release <</ parameters.production_delivery >>
                 name: Bump version
 orbs:
     slack: circleci/slack@3.4.2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -443,6 +443,21 @@ jobs:
                 paths:
                     - ~/.cache/yarn
                     - node_modules
+            - run:
+                command: export GITHUB_RELEASE_NOTES=$(tools/parse-release-notes parse --ios --android --development)
+                name: Parse github release notes
+            - run:
+                command: tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/de-DE/changelogs/default.txt --language de
+                name: Parse german android release notes
+            - run:
+                command: tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/en-GB/changelogs/default.txt --language en && tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/en-US/changelogs/default.txt --language en
+                name: Parse english android release notes
+            - run:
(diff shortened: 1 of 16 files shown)


tools/parse-release-notes
```diff
@@ -0,0 +1,93 @@
...
+        : undefined,
+      ios: ios
+        ? reduceNotes({
+          notes: allNotes.filter(note => note.platforms.includes('ios') && note.platforms.length === 1),
+          language,
+          development
+        })
+        : undefined
+    }
+    return `Release Notes:\n${notesObject.common}\nAndroid:\n${notesObject.android}\niOS:\n${notesObject.ios}`
```
klinzo, line 60: Should we maybe differ (or at least add a note) between things that are shown in the stores because they are relevant to the users and technical and internal stuff that only appears here in the release notes on github? That would probably make it easier for other teams (especially the kommunen/promo team) to find changes that are relevant for them.

.circleci/config.yml
```diff
@@ -443,6 +443,18 @@ jobs:
...
+                name: Parse german android release notes
+            - run:
+                command: tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/en-GB/changelogs/default.txt --language en --production && tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/en-US/changelogs/default.txt --language en --production
+                name: Parse english android release notes
+            - run:
+                command: tools/parse-release-notes parse --ios --destination ios/fastlane/metadata/de-DE/release_notes.txt --language de --production
+                name: Parse german ios release notes
+            - run:
+                command: tools/parse-release-notes parse --ios --destination ios/fastlane/metadata/en-US/release_notes.txt --language en --production
+                name: Parse english ios release notes
```
Taggotty, line 457: didn't you have a fallback to use the default language in the release notes if there are no release notes in the specified language? So what I want to now is whether it can happen that there are e.g two english release nodes or sth.

tools/bump-version
```diff
@@ -11,16 +11,17 @@ program
   .requiredOption('--owner <owner>', 'owner of the current repository, usually "Integreat"')
   .requiredOption('--repo <repo>', 'the current repository, usually "integreat-webapp" or "integreat-react-native-app"')
   .requiredOption('--branch <branch>', 'the current branch')
+  .requiredOption('--release-notes <release-notes>', 'the release notes as stringified string')
```
Taggotty, line 14:
```suggestion
  .requiredOption('--release-notes <release-notes>', 'the release notes as stringified json')
```

Taggotty: or did I misunderstand? ^^

tools/next-version
```diff
@@ -21,7 +21,7 @@ const calculateNewVersion = () => {
   const versionNameCounter = year === versionNameParts[0] && month === versionNameParts[1] ? versionNameParts[2] + 1 : 0
   const newVersionName = `${year}.${month}.${versionNameCounter}`
 
-  if (typeof versionCode !== 'number') {
+  if (versionCode && typeof versionCode !== 'number') {
```
Taggotty, line 24: actually this change is unneccessary, but if you still feel that this is more consitent you can leafe it as it is

tools/parse-release-notes
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  const formattedNotes = notes
+    .map(note => {
+      const localizedNote = note[language] || note[DEFAULT_LANGUAGE]
+      return production ? `* ${localizedNote}` : `* ${note.issue_key}: ${localizedNote}`
+    })
+    .join('\n')
+
+  return platformName ? `\n${platformName}:\n${formattedNotes}` : `${formattedNotes}`
```
Taggotty, line 39:
```suggestion
  return platformName ? `${platformName}:\n${formattedNotes}` : `${formattedNotes}`
```

tools/next-version
klinzo: The aim is to have exactly the same script for native and webapp, so I would prefer to leave it as it is. When we switch to a mono repo, we can use the same script.

tools/parse-release-notes
klinzo: Actually, this is fine :) Otherwise we have `Release notes:ios:` instead of
`Release notes:
ios:`

.circleci/config.yml
klinzo: Not exactly sure if I understand your question correctly. If release notes for e.g. `ar` are created but there are none in that language, it will be completely in english. That should not be a problem though as the fallback/default in the stores is english anyway.

tools/bump-version
klinzo: No, you are right :D

release-notes/ReleaseNoteTemplate.yml
```diff
@@ -0,0 +1,8 @@
+issue_key: <jira issue key> # e.g. NATIVE-587, WEBAPP-42
+show_in_stores: <boolean> # whether this note should be shown in the stores, either true or false
+platforms: # relevant platforms, possible values: android and ios
+  - <platform1>
+  - <platform2>
+en: <your note> # english version of the note
+# optional: the note in other languages:
+de: <your german note> # german version of the note
```
maxammann, line 8: This is different to the format we discussed in some grooming: https://issues.integreat-app.de/browse/WEBAPP-536

Is there a reason for this? I think it makes sense to have multiple items per release note.

.circleci/config.yml
maxammann: We are not sure how release notes are handled by the stores. We will see :shrug:

tools/constants.js
```diff
@@ -0,0 +1,5 @@
+const RELEASE_NOTES_DIR = 'release-notes'
+const UNRELEASED_DIR = 'unreleased'
+const GITKEEP_FILE = '.gitkeep'
+
+module.exports = { RELEASE_NOTES_DIR, UNRELEASED_DIR, GITKEEP_FILE }
```
maxammann, line 5:
```suggestion
export default { RELEASE_NOTES_DIR, UNRELEASED_DIR, GITKEEP_FILE }
```

tools/bump-version
```diff
@@ -11,16 +11,17 @@ program
...
 const commitAndTag = async (versionName, versionCodeString, {
-  deliverinoPrivateKey, owner, repo, branch, developmentRelease
+  deliverinoPrivateKey, owner, repo, branch, developmentRelease, releaseNotes
 }) => {
   const appOctokit = await authenticate({ deliverinoPrivateKey, owner, repo })
   const versionFileContent = await appOctokit.repos.getContent({ owner, repo, path: VERSION_FILE, ref: branch })
 
-  const versionCode = parseInt(versionCodeString)
-  if (isNaN(versionCode)) {
+  const versionCode = versionCodeString ? parseInt(versionCodeString) : null
```
maxammann, line 23: Shouldn't we make versionCodeString required?

.circleci/config.yml
```diff
@@ -443,6 +443,18 @@ jobs:
                 paths:
                     - ~/.cache/yarn
                     - node_modules
+            - run:
+                command: tools/parse-release-notes parse --android --destination android/fastlane/metadata/android/de-DE/changelogs/default.txt --language de --production
```
maxammann, line 447: Should `--production` always be set in this section?

tools/parse-release-notes
```diff
@@ -0,0 +1,126 @@
...
+
+const DEFAULT_NOTE_EN = 'We’ve been working hard behind the scenes to make sure everything is working as it should. If you notice anything that does not work, let us know!\n'
+const DEFAULT_NOTE_DE = 'Wir haben hinter den Kulissen hart gearbeitet, um sicherzustellen, dass alles so funktioniert, wie es soll. Wenn Sie bemerken, dass etwas nicht funktioniert, lassen Sie es uns wissen!\n'
+
+const PLATFORM_ANDROID = 'android'
+const PLATFORM_IOS = 'ios'
+const PLATFORM_WEB = 'web'
+
+program
+  .option('-d, --debug', 'enable extreme logging')
```
maxammann, line 18: This is unused

tools/parse-release-notes
```diff
@@ -0,0 +1,126 @@
...
+
+      // Log to enable bash piping
+      console.log(JSON.stringify(notes))
+    } catch (e) {
+      console.error(e)
+      process.exit(1)
+    }
+  })
+
+program.parse(process.argv)
```
maxammann, line 126: Not reviewed as it seems like in NATIVE

maxammann (approved):

release-notes/ReleaseNoteTemplate.yml
klinzo: Yep, right. I first wanted to use that format but later figured this just makes more sense. My reasons:
* If we have multiple items, we need to define whether it is a production note or not for each item separately which makes the format more complicated than the one discussed in the grooming.
* Normally there should not be more than one note per issue, but if there are more than one, one can simply add multiple release note files.
* This format is simpler than the one discussed and therefore easier to use for developers.

tools/bump-version
klinzo: See discussion here: https://github.com/Integreat/integreat-react-native-app/pull/286#discussion_r475562584

.circleci/config.yml
klinzo: Yes with the exception of release notes for github.

.circleci/config.yml
klinzo: Yes :D

tools/parse-release-notes
klinzo: Yes, it is the same like in the webapp.

maxammann:
ready for merge :)

PR #294, opened 2020-08-11
NATIVE-592: App crashes on ios

klinzo:


ios/Integreat.xcodeproj/xcshareddata/xcschemes/integreat.xcscheme
```diff
@@ -89,7 +89,7 @@
          <BuildableReference
             BuildableIdentifier = "primary"
             BlueprintIdentifier = "13B07F861A680F5B00A75B9A"
-            BuildableName = "Integreat.app"
+            BuildableName = "IntegreatE2E.app"
```
maxammann, line 92: Can you change this back to Integreat.app? I think until we fixed this we should make sure not to commit these changes

PR #297, opened 2020-08-12
NATIVE-595 + NATIVE-599 + NATIVE-593: Add firebase config for malte

maxammann:


ios/Integreat/AppDelegate.m
```diff
@@ -33,15 +33,18 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
       options.storageBucket = @BUILD_CONFIG_IOS_GOOGLE_SERVICES_STORAGE_BUCKET;
+      
+      // The following values are unset as they are probably not relevant for Messsaging:
+      // options.deepLinkURLScheme = ??
+      // options.trackingID = ??
       // options.appGroupID = ??
       
-      // Unused info from GoogleService-Info.plist
+      // Unused info from GoogleService-Info.plist:
+      // These exist in the .plist file but I was not able to add them to the FIROptions. They are probably not important.
```
klinzo, line 47:
```suggestion
      // These existed in the .plist file but I was not able to add them to the FIROptions. They are probably not important.
```

Also: Couldn't we remove them from the build config then?

maxammann: I would leave them here as it is unclear what effects this has. The documentation is not that good.

klinzo (approved):

PR #298, opened 2020-08-12
NATIVE-595 + NATIVE-599: Use proper ios certificates

maxammann:


build-configs/configs/malte/index.js
```diff
@@ -28,7 +28,8 @@ const MalteBuildConfig: BuildConfigType = {
     }
   },
   ios: {
-    bundleIdentifier: 'de.integreat-app',
+    bundleIdentifier: 'de.malteapp',
+    provisioningProfileSpecifier: 'match Development de.malteapp',
```
klinzo, line 32: Do we really need this? Wouldn't it be enough to just use `match Development ${bundleIdentifier}`?

klinzo: I'd try to avoid having redundant information here if possible.

maxammann: This option makes sense if you want to use a AdHoc or AppStore profile. Then you can quickly set this in the build config.

fastlane/Fastfile
```diff
@@ -88,7 +88,7 @@ platform :android do
     app_name = options[:app_name]
 
     generic_sentry_upload(
-        package_name: "tuerantuer.app.integreat", # FIXME
+        package_name: "tuerantuer.app.integreat", # FIXME: bundle id
```
Taggotty, line 91: why is there the fixme for the bundle id? do we only get this id if we have had our "first" release?

Taggotty (approved):

fastlane/Fastfile
maxammann: I'm currently resolving this in my current task. We already know the id, but yes, this will only be used as soon as we do the first release.

PR #300, opened 2020-08-17
NATIVE-597: Configurable icons

klinzo:


src/modules/app/constants/buildConfig.js
```diff
@@ -2,7 +2,32 @@
...
 import loadBuildConfig from '../../../../build-configs'
+import { INTEGREAT_ICONS } from '../../../../build-configs/configs/integreat'
+import { MALTE_ICONS } from '../../../../build-configs/configs/malte'
+import integreatAppLogo from '../../../../build-configs/configs/integreat/assets/app-logo.png'
+import integreatLocationMarker from '../../../../build-configs/configs/integreat/assets/location-marker.svg'
+import malteAppLogo from '../../../../build-configs/configs/malte/assets/app-logo.svg'
+import malteLocationMarker from '../../../../build-configs/configs/malte/assets/location-marker.svg'
 
-const buildConfig = (): BuildConfigType => loadBuildConfig(process.env.BUILD_CONFIG_NAME)
+const buildConfigIcons = (iconSet: string) => {
```
Taggotty, line 12: I don't quite get why you parse the itemset here instead of directly parsing th MalteBuildConfig. Is it because of the test maybe? Or is there any other reason?

klinzo: You mean why I don't include the icons directly in the build config? Importing .svg files does not work with plain js. Also, we would need to handle .svgs and .pngs separately upon creating xcconfig files since these are simple key/value files. I tried it anyway, but couln't get it to work, maybe because the svg plugin we are using (https://github.com/aeirola/react-native-svg-asset-plugin) does not work for files outside the src dir.

klinzo: Any ideas would be appreciated though :)

Taggotty: ah. okay. that's annoying ^^ . Then I'd say leave it as it is, even though I still feel it is kind of out of place here. Maybe @maxammann  has a plan

klinzo: Yep, maybe also @Schedulaar ?

src/modules/app/constants/buildConfig.js
```diff
@@ -2,7 +2,32 @@
...
+  if (iconSet === INTEGREAT_ICONS) {
```
maxammann, line 13: You can also check here directly for the appName. There is no need to a separate build config option if we handle it like this imo.

src/modules/app/constants/buildConfig.js
maxammann: Something that should work according to the README of the react-native-svg-asset-plugin is to use the `require` call.

So include the following in the build config:
```
iconPaths: {
  appLogo: './build-configs/configs/integreat/assets/app-logo.png' // path relative from project root
}
```

Then in the code you can use `require('../../../../' + buildConfig.iconPaths.appLogo)`

maxammann: The Build Config only can contain primives: bools, arrays, maps, strings, numbers etc.
Also the config should only import js files as it needs to work with only nodejs.

An improvment to that way would be to have a function: `loadIcons(buildConfig)` which returns an object of paths to the PNGs of react-native-svg-asset-plugin.

src/modules/app/constants/buildConfig.js
```diff
@@ -2,7 +2,32 @@
...
+  } else if (iconSet === MALTE_ICONS) {
+    return {
+      appLogo: malteAppLogo,
+      locationMarker: malteLocationMarker
+    }
+  }
+  throw new Error(`Unknown icon set ${iconSet}. Check your build config!`)
+}
+
+const buildConfig = (): {| ...BuildConfigType, appLogo: string, locationMarker: string |} => {
```
maxammann, line 27: Returning here other data together with the BuildConfigType seems very :/

src/modules/app/constants/buildConfig.js
Taggotty: Not sure if that is actually "prettier" than using the flag. Surely it is more extensible, but handling Icons feels all over the place like this...

src/modules/app/constants/buildConfig.js
klinzo: integreat-e2e and integreat-test-cms are using the same icons as integreat but have different app names.

maxammann:
Apart from the design question this is ready for merging :)

src/modules/app/constants/buildConfig.js
klinzo: As discussed the suggestion by @maxammann does not work with the plugin. Leaving it as it is.

src/modules/app/constants/buildConfig.js
klinzo: Should I create a separate function to do this?

klinzo:
@maxammann do you mean this comment https://github.com/Integreat/integreat-react-native-app/pull/300#discussion_r475556685 by this?
> Apart from the design question this is ready for merging :)

src/modules/app/constants/buildConfig.js
maxammann: I think that would be good, or rename to buildConfigWithIcons

maxammann (approved):

maxammann (approved):

PR #299, opened 2020-08-17
NATIVE-604: Add release notes docu

klinzo:
This pull request belongs to an issue on our [bugtracker](https://issues.integreat-app.de/).
You can find it there by looking for an issue with the key which is mentioned in the title of this pull request.
It starts with the keyword **NATIVE**.

PR #302, opened 2020-08-19
NATIVE-609: Migrate to main

klinzo:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -514,7 +514,7 @@ workflows:
                 filters:
                     branches:
                         only:
-                            - develop
+                            - main
     bi_weekly_production_delivery:
         jobs:
             - prepare_build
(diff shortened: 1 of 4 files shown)


maxammann (approved):
You will create the main branch when merging this, right?

maxammann:
I think this should work fine :)

Taggotty:
Still found one occurence of develop in Contributing.md . Please checkout if it should be replaced. 
I am a little bit worried that the bump-version will not work as expected whith a new branchref, but I guess we will see :)

Taggotty (approved):

klinzo:
@Taggotty good catch, the line will be removed in https://github.com/Integreat/integreat-react-native-app/pull/299/files anyway, so I won't replace it here.

Not sure why it should not work as we get the new branch ref, so I hope you are wrong. If not though, I stand corrected :D

PR #319, opened 2020-09-14
NATIVE-384: Show correct city when linking internally

maxammann:

### DIFF
--- a/src/modules/layout/containers/HeaderContainer.js
+++ b/src/modules/layout/containers/HeaderContainer.js
@@ -27,7 +27,7 @@ type StatePropsType = {|
   goToLanguageChange?: () => void,
   peeking: boolean,
   categoriesAvailable: boolean,
-  cityModel?: CityModel
+  routeCityModel?: CityModel
 |}
 
 type DispatchPropsType = {|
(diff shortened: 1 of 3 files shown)


src/modules/layout/containers/HeaderContainer.js
```diff
@@ -47,13 +47,11 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
   const cityCode = state.cityContent?.city
   const categoriesAvailable = state.cityContent?.searchRoute !== null
 
-  const cityModel = cities.find(city => city.code === cityCode)
```
Schedulaar, line 50: You probably could have instantiated `cityModel` here already:
```js
const cityModel = route ? cities.find(city => city.code === route.cityCode) : null
```
in order to pass it in the first step already when other things are still loading...

src/modules/layout/containers/HeaderContainer.js
```diff
@@ -47,13 +47,11 @@ const mapStateToProps = (state: StateType, ownProps: OwnPropsType): StatePropsTy
   const cityCode = state.cityContent?.city
```
Schedulaar, line 47: Maybe rename this variable to `stateCityCode` or similar...

Schedulaar (approved):

klinzo (approved):

klinzo (commented):
Please add a release note :)

PR #321, opened 2020-09-14
NATIVE-622: Fix Sentry error reports

maxammann:


build-configs/configs/BuildConfigType.js
```diff
@@ -49,7 +49,8 @@ export type BuildConfigType = {|
   lightTheme: ThemeType,
   darkTheme: ThemeType,
   iconSet: string,
-  development?: boolean,
+  development: boolean,
+  sentry: boolean,
```
klinzo, line 53: Shuldn't this be a part of featureFlags?

klinzo (approved):
What was the problem? And how does this fix it?

build-configs/configs/BuildConfigType.js
klinzo: Still relevant

maxammann: done

klinzo (approved):

PR #323, opened 2020-09-15
NATIVE-601: Fix push notification settings

klinzo:

### DIFF
--- a/src/routes/settings/components/Settings.js
+++ b/src/routes/settings/components/Settings.js
@@ -15,8 +15,8 @@ import type { SectionBase } from 'react-native/Libraries/Lists/SectionList'
 
 type PropsType = {|
   theme: ThemeType,
-  language: string,
-  city: ?string,
+  languageCode: string,
+  cityCode: ?string,
   t: TFunction,
   navigation: NavigationScreenProp<*>,
   dispatch: () => {}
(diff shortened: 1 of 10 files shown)


klinzo (commented):
Sentry is not yet stopped/started instantly if a setting is changed, only on the next app start.

src/routes/settings/components/Settings.js
```diff
@@ -10,13 +10,13 @@ import type { TFunction } from 'react-i18next'
...
 import type { SettingsType } from '../../../modules/settings/AppSettings'
 import createSettingsSections from '../createSettingsSections'
-import type { ChangeSettingFunctionType } from '../createSettingsSections'
 import AppSettings, { defaultSettings } from '../../../modules/settings/AppSettings'
 import type { SectionBase } from 'react-native/Libraries/Lists/SectionList'
 
 type PropsType = {|
   theme: ThemeType,
   language: string,
+  city: ?string,
```
maxammann, line 19: Probably `cityCode` is better here

maxammann (approved):

src/routes/settings/components/Settings.js
klinzo: Yes, kind of. We are really inconsistent with this. E.g. language here is also rather a languageCode than a language.

Taggotty (approved):

maxammann (approved):

PR #328, opened 2020-09-16
NATIVE-632: Prevent firebase not initialized error

klinzo:


src/modules/endpoint/sagas/loadCityContent.js
```diff
@@ -38,7 +38,9 @@ export default function * loadCityContent (
 
   if (!criterion.peeking()) {
     const appSettings = new AppSettings()
-    yield call(() => NotificationsManager.subscribeNews(newCity, newLanguage))
+    if (buildConfig().featureFlags.pushNotifications) {
```
maxammann, line 41: I think it would be cleaner to handle this check within the NotificationsManager.
That way you can always call subscribeNews but it will have no effect if PushNotifications are disabled. 

Maybe NotificationsManager should be renamed to PushNotificationsManager or similar.

src/routes/intro/IntroContainer.js
```diff
@@ -171,7 +171,7 @@ class Intro extends React.Component<PropsType, StateType> {
         await requestLocationPermission()
       }
 
-      if (allowPushNotifications) {
+      if (allowPushNotifications && buildConfig().featureFlags.pushNotifications) {
         await requestPushNotificationPermission()
```
maxammann, line 175: Maybe requestPushNotificationPermission could also be moved to NotificationsManager. Not sure though

maxammann (approved):

maxammann (approved):

build-configs/index.js
```diff
@@ -21,6 +21,15 @@ const loadBuildConfig = (buildConfigName: ?string): BuildConfigType => {
   if (!buildConfig) {
     throw Error(`Invalid BUILD_CONFIG_NAME supplied: ${buildConfigName}`)
   }
+
+  const { featureFlags: { pushNotifications }, android, ios } = buildConfig
+
+  if (android.googleServices === null !== ios.googleServices === null) {
+    console.warn('Google services should be configured for both or no platform!')
+  } else if (pushNotifications === (android.googleServices === null)) {
+    console.warn('Push notification feature flag does not match google service configuration!')
```
Schedulaar, line 30: Probably errors might be better here... Would be even better to check this at compile time (or is this compile time^^).

klinzo: Not sure... If I add just add a new firebase configuration but don't want to have the feature enabled already this is fine imo.

src/modules/app/LocationPermissionManager.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+import { check, PERMISSIONS, request, RESULTS } from 'react-native-permissions'
+import { Platform } from 'react-native'
+
+export const locationPermissionStatus = async (): Promise<RESULTS> => {
```
Schedulaar, line 6:
```suggestion
export const checkLocationPermission = async (): Promise<RESULTS> => {
```

Taggotty (approved):

Schedulaar (approved):

lulo4907 (approved):

PR #331, opened 2020-09-17
NATIVE-623: Windows OS gradle build error xcconfig

lulo4907:


klinzo:
Aren't these exactly the same changes as in the other PR? :thinking: Maybe you should choose the other branch as target of the PR.

maxammann (approved):

lulo4907:
@klinzo This PR is technically not related to the previous PR. The task occurred in the wake of a previous one. Therefore I thought created a separate entry.

android/app/buildConfigs.gradle
```diff
@@ -13,7 +24,7 @@ def createBuildConfig(buildConfigName) {
     logger.quiet("Using build config $buildConfigName")
 
     def command = "yarn --silent babel-node tools/build-config to-json $buildConfigName"
-    def proc = command.execute()
+    def proc = execCommand(command)
     proc.waitFor()
```
klinzo, line 28: If it works only waitFor, it should be removed here as well.

android/app/buildConfigs.gradle
```diff
@@ -48,16 +59,13 @@ def setupGoogleServices(buildConfig, resValue) {
...
+
     def command = "yarn --silent babel-node tools/build-config to-xcconfig $buildConfigName"
-    def proc = command.execute()
-    proc.waitFor()
+    def proc = execCommand(command)
 
-    if (proc.exitValue() != 0) {
-        println("Failed to get xcconfig file: $proc.err.text")
-        return
-    }
```
klinzo, line 58: Why did you remove this? This is kind of important for debugging if the execution of the command fails.

android/app/buildConfigs.gradle
```diff
@@ -48,16 +59,13 @@ def setupGoogleServices(buildConfig, resValue) {
 }
 
 def setupResourceValues(buildConfigName, resValue) {
+
```
klinzo, line 62: Remove newline

steffenkleinle (changes requested):

android/app/buildConfigs.gradle
lulo4907: will be done

android/app/buildConfigs.gradle
lulo4907: will be done

android/app/buildConfigs.gradle
lulo4907: This has caused the last CI failure. Re-entered with re-arrangement as next.

klinzo: Ahh yes I guess I know what the problem is now. `proc.in.text` waits until the process is finished and the input stream is closed. If you call `proc.exitValue()` before that and without `proc.waitFor()`, the process is not yet finished and fails therefore. So you probably have to move this block after `proc.in.text`.

android/app/buildConfigs.gradle
lulo4907: The waitFor is necessary in this case, otherwise it runs into a "not exited" error.

android/app/buildConfigs.gradle
klinzo: Still relevant.

klinzo: Still relevant :)

release-notes/unreleased/NATIVE-632-WindowsOs-gradle-build-error-xcconfig.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: NATIVE-623
```
klinzo, line 1: Imo a release note is not necessary and even does not make sense since it makes the release notes bigger without benefit.

android/app/buildConfigs.gradle
klinzo: See https://github.com/Integreat/integreat-react-native-app/pull/331/files#r492143073. The check for `proc.exitValue()` has to be after `proc.in.text`.

steffenkleinle (changes requested):

android/app/buildConfigs.gradle
lulo4907: At my tests changing this workflow results in an process error at CI.

sarahsporck (approved):
tested on my pc and got it running on windows 👍🏼

steffenkleinle (approved):

steffenkleinle (approved):

PR #332, opened 2020-09-18
NATIVE-577: Show matching category content in search

sarahsporck:


maxammann:
I fixed the title of this PR such that it matches the others.

src/modules/categories/components/CategoryList.js
```diff
@@ -9,9 +9,16 @@ import styled from 'styled-components/native'
 import Image from '../../common/components/Image'
 import CategoryListCaption from '../../../modules/common/components/CategoryListCaption'
 
+export type SimpleModelType = {|
```
maxammann, line 12: Maybe `ListElementModelType` is a better name as there is no complex or advanced ModelType

src/modules/categories/components/ContentMatcher.js
```diff
@@ -0,0 +1,48 @@
+// @flow
```
maxammann, line 1: ContentMatcher is no component. Therefore it should be moved one level up.

src/routes/search/components/SearchModal.js
```diff
@@ -51,27 +52,45 @@ class SearchModal extends React.Component<PropsType, StateType> {
...
+    const parser = new Parser({ ontext (data: string) { contentWithoutHtml.push(data) } })
     const categoriesWithContent = categoriesArray
-      .filter(category => !normalizeSearchString(category.title).includes(normalizedFilter))
-      .filter(category => normalizeSearchString(category.content).includes(normalizedFilter))
-      .sort((category1, category2) => category1.title.localeCompare(category2.title))
+      .filter(category => !normalizeSearchString(category.title).includes(normalizedFilter) && !category.isRoot())
+      .map((category: CategoryModel): ListEntryType => {
+        contentWithoutHtml = []
+        parser.write(category.content)
+        parser.end()
```
maxammann, line 76: If you call parser.end() within the map function then I would assume that the parser also gets initialized here.

maxammann (approved):

release-notes/unreleased/NATIVE-577-CategorySearchWithContent.yml
```diff
@@ -0,0 +1,7 @@
+issue_key: NATIVE-577
+show_in_stores: true
+platforms:
+  - android
+  - ios
+en: The search now also highlights the content of a category it matches.
```
klinzo, line 6:
```suggestion
en: The search now also highlights the matching category content.
```

release-notes/unreleased/NATIVE-577-CategorySearchWithContent.yml
```diff
@@ -0,0 +1,7 @@
...
+de: Die Suche hebt nun auch den Inhalt hervor, wo sie mit dem Filter übereinstimmt.
```
klinzo, line 7:
```suggestion
de: Die Suche hebt nun auch den Inhalt hervor, der mit dem Filter übereinstimmt.
```

src/modules/categories/components/CategoryList.js
klinzo: Kind of... its a simpler CategoryType with additional `contentWithoutHtml`, but I'd also vote for something different. I don't think `Element` and `Model` are necessary, maybe simply `CategoryListModelType?

src/modules/categories/components/ContentMatcher.js
klinzo: We also usually don't have classes for functionality, do we? Of course this is also okay, just not sure if a class makes sense here.

src/modules/categories/components/CategoryListItem.js
```diff
@@ -27,16 +31,17 @@ const DirectionContainer: StyledComponent<DirectionContainerPropsType, ThemeType
   flex-direction: ${props => contentDirection(props.language)};
 `
 
-const CategoryTitleContainer: StyledComponent<DirectionContainerPropsType, ThemeType, *> = styled.View`
+const CategoryEntryContainer: StyledComponent<DirectionContainerPropsType, ThemeType, *> = styled.View`
   flex: 1;
-  flex-direction: ${props => contentDirection(props.language)};
+  flex-direction: column;
```
klinzo, line 36: Why did you change this and moved it to `CategoryTitle`?

src/modules/categories/components/CategoryListItem.js
```diff
@@ -78,13 +85,33 @@ class CategoryListItem extends React.Component<PropsType> {
     )
   }
 
+  getMatchedContent (numWordsSurrounding: number): ?Highlighter {
+    const { query, theme, category } = this.props
+    const textToHighlight = this.contentMatcher.getMatchedContent(query, category.contentWithoutHtml, numWordsSurrounding)
```
klinzo, line 90: Any reason why you have to initialize contentMatcher for the whole component?

src/modules/categories/components/CategoryListItem.js
klinzo: Especially, is the matched text in the right direction?

src/modules/categories/components/ContentMatcher.js
klinzo: This could also not really category specific but could be used for events, ... Maybe it should even live somewhere else, idk.

src/routes/search/components/SearchModal.js
```diff
@@ -51,27 +52,45 @@ class SearchModal extends React.Component<PropsType, StateType> {
...
+          path: category.path
+        },
+        subCategories: []
+      })
+
+      )
+      .sort((category1, category2) => category1.model.title.localeCompare(category2.model.title))
 
     // find all categories whose contents but not titles include the filter text and sort them lexicographically
+    let contentWithoutHtml = []
```
klinzo, line 69: Can't we inline this and convert to const?

src/modules/categories/components/ContentMatcher.js
maxammann: Yeah the function could just be exported as `export const ...`

lulo4907 (approved):

src/modules/categories/components/CategoryListItem.js
Taggotty: The main reason why I did it, is because that is what I did in the webapp and nobody complained. Furthermore it kind of feels incorrect to reinstantiate the contentMatcher on every render

src/modules/categories/components/ContentMatcher.js
Taggotty: I think it was @Schedulaar who recommended using a separate class for the ContentMatcher in the webapp, as it simplifies testing. 



> Yeah the function could just be exported as `export const ...`

Do you mean that I scrap the class and keep the functions as local functions? My question then would be, whether it is possible to to test the functions the and whether I should apply those changes to the webapp.

src/routes/search/components/SearchModal.js
Taggotty: I am kind of worried that the Parser initialization will impact the performance at least visibly. But I will test it shortly.

Taggotty: Seems to not matter

steffenkleinle (approved):

